PR #2, opened 2016-12-18
Add ResponseUtil to the web/util package

gzsombor:
The new util class


jdubois:
We should find a better name than 'ResponseUtil' but it's just too late for me to find a good one :-)

gzsombor:
Yep :) We already have HeaderUtil and PaginationUtil, so I thought this is good enough :)

jdubois:
Oh yes you are correct...

PR #5, opened 2017-01-23
add the jhipster.registry.password property

PierreBesson:
It will require a new release of the lib. Can't we automate releasing with Travis ?
cc @pascalgrimaud


jdubois:
Nope because that's released to Maven Central, and you can't have snapshots there. Don't worry I'll do a release.

pascalgrimaud:
@PierreBesson : yes it's possible. You can do what you want with Travis... but as Julien said, you have to write a script to change SNAPSHOT, or use maven release plugin... there are some work to do, adding keys etc.

So if Julien is not too tired, manually release is OK for me, then maybe one day, we can begin to automate some workflows ;-)

PR #9, opened 2017-02-28
Map ByteBuffer to string in swagger

cbornet:


toolforger:
I'm wondering whether this is going to be a perpetually increasing list of converters.
Maybe if you said something about what problem is being solved here?

cbornet:
ByteBuffer is a type available for Cassandra

gmarziou:
@toolforger this is just to have a more readable API documentation, it does not have any impact on API itself so I don't see any problem adding types that are common.

PR #11, opened 2017-03-15
Consistent log messages for AsyncSpringLiquibase

toolforger:
"Starting liquibase" without an associated "stopping liquibase" message made me think that Liquibase is kept running, which does not make sense and indeed is not what's happening.
I took the liberty of giving the process the "DB setup via Liquibase" name, and use that consistently in log messages so readers can associate the message with a process without having to look at the (possibly abbreviated) class name.

I also removed the `e.getMessage()` call:
* by default, the exception message will be printed by logger.error(..., e) anyway so it's just going to be noise;
* if an installation is configured to not show exception messages, `AsyncSpringLiquibase` shouldn't subvert that decision IMHO. E.g. logback's default formatter allows "suppress all exception info", "show just the exception message", and "show message and full call stack".

### DIFF
--- a/src/main/java/io/github/jhipster/config/liquibase/AsyncSpringLiquibase.java
+++ b/src/main/java/io/github/jhipster/config/liquibase/AsyncSpringLiquibase.java
@@ -57,19 +57,18 @@ public class AsyncSpringLiquibase extends SpringLiquibase {
                 .SPRING_PROFILE_HEROKU)) {
                 taskExecutor.execute(() -> {
                     try {
-                        logger.warn("Starting Liquibase asynchronously, your database might not be ready at startup!");
+                        logger.warn("Scheduling background DB setup via Liquibase; the database will not be ready until DB setup finishes");
                         initDb();
                     } catch (LiquibaseException e) {
-                        logger.error("Liquibase could not start correctly, your database is NOT ready: {}", e
-                            .getMessage(), e);
+                        logger.error("Error scheduling DB setup via Liquibase; the database will not become ready, reason:", e);
                     }
                 });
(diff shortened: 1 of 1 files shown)


jdubois:
I'm not a big fan of "DB setup", but yes you've got a good point, this needs to be changed.

toolforger:
I didn't think much about the terminology, so "DB setup" was what I came up with.
I see that this terminology is a bit fuzzy, but I am not sure what exactly should be expressed here, so I am asking for feedback.
* A better term would be fine.
* Opinions about *what* should be expressed would move this ahead too, in that case I'd try to come up with a better term myself.

jdubois:
Reading this again:

- For the log message, this is because one is a stacktrace, and the other is not, so they might not end up in the same place. As this is a fatal error, I'd rather have this logged twice.
- For the text, I'm happy with "Liquibase", but indeed the last message isn't very good

-> I'm going to rewrite the last message and point my change to this ticket.

That's why in the guidelines you should first open up a ticket, so we can discuss the change, before doing a PR.

jdubois:
Oh, and just an idea (which I probably won't time to do), maybe we could name the thread specifically. In that case we wouldn't take it in the thread pool (very minor perf issue), but at least it could easily be monitored.

jdubois:
Sorry for my last comment: our thread pool already has a specific name, so let's not create too many of those!

jdubois:
Closing this as I have changed the message in https://github.com/jhipster/jhipster/commit/4a4a987950373e0caf8dfe09bf31530a2f41a894

Thanks for pointing this out, we should discuss this more next time

PR #14, opened 2017-04-26
Align versions with generator

erikkemperman:
Incidentally, I have been unwell for a while but now back to working toward an alpha / proof-of-concept of my jhipster-dependencies proposal. I am considering taking this module along as well which would mean this kind of alignment would become automatic / implicit.


erikkemperman:
Thanks Julien!

PR #17, opened 2017-06-08
Infinispan - initial commit - add attributes

srinivasa-vasu:


src/main/java/io/github/jhipster/config/JHipsterProperties.java
```diff
@@ -191,6 +193,8 @@ public Ehcache getEhcache() {
             return ehcache;
         }
 
+        public Infinispan getInfinispan() { return infinispan; }
```
MathieuAA: It would be better to have this getter this way instead of a one-liner (take `getEhcache` for instance)
```java
public Infinispan getInfinispan() {
  return infinispan;
}
```

srinivasa-vasu: Done

src/main/java/io/github/jhipster/config/JHipsterProperties.java
```diff
@@ -236,6 +242,81 @@ public void setMaxEntries(long maxEntries) {
...
+
+            private String configFile = "default-configs/default-jgroups-tcp.xml";
+
+            private final Local local = new Local();
+
+            private final Distributed distributed = new Distributed();
+
+            private final Replicated replicated = new Replicated();
+
+            public String getConfigFile() { return configFile; }
```
deepu105: @srinivasa-vasu could you plz format the new code similar to remaining of the file so that getter/setters arent on the same line to be consistent with codestyle of the file

srinivasa-vasu: My bad, I thought I changed it already.

PR #21, opened 2017-07-09
generator-jhipster#6061 First step: change the project structure in order to further add pom hierarchy

Tcharl:
Changes the hierarchy of the project to be able to add new modules.
There will be new pom module hierarchy:
Some vendor neutral ones (for reporting, plugin management, dependency management) and one vendor specific (the jhipster corp one, containing the list of commiters, licence, Etc).

Change-Id: Idb619e8962422a91183295fec7ec268ed84ee6ec
Signed-off-by: charlie <cmordant1@gmail.com>


jdubois:
I thought we were just doing a pom that could be imported? I don't understand why we change the project hierarchy here.
This is also going to cause issues with the release mechanism on Maven Central, I don't think we should do this. We should have one GitHub project per Maven project.

Tcharl:
So you have to initiate a new jhipster-parent repo: the jhipster/jhipster repo contains a jar project. If we want to have a pom project and a jar project, we've to make a parent pom project and modules like I did. So if you want that repo hierarchy to be left untouched and have a new pom project, we should host it on another git repo

Tcharl:
Note that a mvn release:perform won't change anything with this layout: the jar will have the same group and artifactIds, only a new pom project will be released (jhipster-parent)

Tcharl:
I Just noticed that you don't use the maven release plugin(no scm section, no -SNAPSHOT module version on the master).
Do you want to have that plugin support? it will ease the deployement, creating tags automatically, incrementing the version number, Etc.

jdubois:
- oh yes you are correct, we needed a jhipster-parent repo, and I need to do so paperwork to have it on Maven Central
- for deploying: thanks, if it's really easy to use. I have no real issue today, it's just a couple of simple commands, but of course if that can be automated it's better.

Tcharl:
FYI, I have initiated the pom structure [here](https://github.com/Tcharl/jhipster-parent)

Tcharl:
closing this PR in favor of https://github.com/jhipster/jhipster/pull/22

PR #22, opened 2017-07-09
generator-jhipster#6061 Adaptation to the new jhipster-corp pom

Tcharl:
I adapted the project using the 'pom framework' here https://github.com/Tcharl/jhipster-parent

Feel free to create the jhipster-parent on the jhipster organization side and use it for the jhipster corp project.

Features:

- A jhipster 'corporate pom' to fill developpers, ci info, license...

- Automated ACL V2 header generation for corporate projects (using the maven-licence plugin

- Use of the maven-release-plugin to release projects to the maven central (tagging automatically the scm, deploying to sonatype staging, see readme.md for infos)

- Maven reporting standard site generation (goal: site site:stage will generate the maven site on /tmp/deploy)

- A bom to manage project dependencies (tested for the _jhipster_ project: we should maybe add more)

- Plugin management for plugins versions and behavior

- Vendor neutral pom hierarchy (only the child jhipster-corp pom contains jhipster specific information)

- Apache compliance checking before jhipster project release.


PierreBesson:
I would prefer if it was called jhipster-org rather than jhipster-corp. This is the term we use on github.

erikkemperman:
Nice work! Unfortunately it does not seem to solve the "problem" I wanted to address in my earlier issue: there is still a parent dependency on spring-boot-starter-parent (I was looking for open-ended parent hierarchy so I can attach my own) and if I read this correctly it also does not allow the gradle build to use automatically the same dependencies from the maven build.

Sorry, I don't mean to sound harsh and of course I'm looking at this slightly selfishly from my own use-case ;-)

deepu105:
Guys keep just one thing in mind, all this shouldn't make using JHipster harder than it is today for our major user base unless the value is much much greater. Please remember the needs of the many outweighs the needs of the few

erikkemperman:
@deepu105 Definitely! I _think_ that what I have in mind would make things significantly easier, actually, for both users and devs... I will try to find some time this week to submit something concrete. Hopefully we could merge @Tcharl's effort with mine for a result that will be satisfactory all around!

Tcharl:
@PierreBesson I will fix that ;-).

@deepu105 I don't think this solution is harder than the actual: the end user's pom size is dramatically reduced (so its maintenance too), also, the end user's artifact release management (in an internal repo, or in maven central) will be simplified, using the 'standard' way of maven to manage releases. Finally, static indicators site is a bonus for the ones which see some added value in it (for example, the ones who want to control dependencies we were talking about [in the original issue](https://github.com/jhipster/generator-jhipster/issues/6061).

@erikkemperman it should be very nice to be able to reuse dependencies list within gradle. 
Also, multimodule architecture could have its advantages in a 'microservice' mode (having a .mvnw at the root that starts/stops the microservices in the right order!!), but I think that it would be overkill in a 'monolith' mode. 

In a first time, the quick way to answer to your (selfish ;-)) use case is to only copy the pluginManagement and properties sections of the jhipster-plugins pom in your generated pom and use the bom as usual: I could make a PR on generator-jhipster (with an option) to generate this in the standalone pom.xml (it will just be the actual pom generation but using bom, filtering some data in case of superpom usage).

In the end, one way to answer for your  (selfish ;-)) use case and this one would be to provide options on the jhipster-generator in case of a maven+monolith. For example: use the standard jhipster parent pom (providing some cool stuff for free and reducing the amount of config)?yes|no=>would you like to generate a multi-module project adding this stuff?yes|no.

There are some improvement that could be done to this work, and some PRs to do in the jhipster-generator repo (for example, an option to generate a correct site.xml file with fluido skin and jhipster svg).

deepu105:
I was just making sure you guys understood our goal. Also please note adding more questions for this is not preferred and personally I will be strongly against it. I don't think the value is justified enough for adding a new question. If its better than current setting then it should just replace it and become the default

Tcharl:
@deepu105 ah, ok, I wasn't aware about this, but you are right (as always ;-)), it is the good way to go.

So basically, jhipster org have six choices with this PR:

- Keeping things as they are (I would make a PR to beautify the actual pom if its Okay), or adding the content (or a part) of the non-corp poms within the actual template.

- Using the Bom to manage dependency versions: it will ease generated pom maintenance, but, as tienq said, will shadow some update parts to the end user.

- Using the Jhipster-plugins pom as a parent for each generated pom: it will ease pom maintenance (more than the one before, as it also contains the pluginmanagement section), provide some more accelerators (release management, static site generation, license header generation, ...), ease development maintenance (we're not working on a template for poms). The drawback is for those which have corporate poms (but IMHO, the problem is already here, as the actual pom generation inherits from spring-boot).

- Generating a multi module maven project containing the poms' code (I think that this is the @erikkemperman solution if I'm not wrong).

- Make a module (but it's kinda hard to maintain, a new jhipster release with a dependency update would mean a new jhipster-parent pom release).

It's up to you and the jhipster conclave, for my projects (personal or within my actual company), I would choose the 3rd (and this PR is all about it ;-)).

erikkemperman:
@Tcharl Well, not "generating a multi-module maven project", or at least not from the user's point of view.

There would be a parent POM with the jhipster generic properties, dependencies (imported from a separate , plugins but it would be in the same directory structure as the main (user-visible) POM. So the parent is resolved relatively (e.g. `./parent/pom.xml`), and is tailor-made for the project by the user's JH configuration, just like the main POM.

And, crucially, this parent should _not_ have another parent -- so that I can easily inherit from my own "corp/org" parent at this point, or insert an intermediate parent common to all microservices within a project.

I'm sorry I seem to be unable to articulate properly what I mean exactly... Better to show some code, probably, but I need to find some time for it :-)

sendilkumarn:
Good one @Tcharl 

I kinda like this idea.

pom.xml
```diff
@@ -1,371 +1,260 @@
...
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
 
-    <parent>
-        <artifactId>spring-boot-starter-parent</artifactId>
-        <groupId>org.springframework.boot</groupId>
-        <version>1.5.4.RELEASE</version>
-        <relativePath/>
-    </parent>
+	<parent>
```
deepu105: formatting seems weird

Tcharl: I will fix that too: kinda strange because it is the default eclipse one (fresh Oxygen install used only for the purpose).

Tcharl:
@erikkemperman , got it! Is it well handled by IDE like Eclipse? i.e. not recognized as project but handled correctly by ugly maven overrides like m2e? Would you split the generated poms like I've done (three poms: parent, reporting and plugins) or only generating a big big one?

If so, it's kinda nice idea: I've never seen this pattern, but it looks good ;-). But I don't see how we get rid of that spring-boot-starter-parent in all case. So IMHO, maintaining the parent poms outside the client code stays better, because it reduces the amount of code to maintain on the client side and ease the maintenance by the team, but prevents linking it to a corporate pom. In fact, it depends on the use cases, so it's up to the jhipster team (which holds the statistics) to choose: let implement the two solutions ;-).

erikkemperman:
@Tcharl I don't expect IDEs will find this setup problematic.

And spring-boot-starter-parent doesn't do all that much, mostly it imports spring-boot-dependencies.

My idea was we could import from a new module, jhipster-dependencies, extending spring-boot-dependencies.

 Then we will basically have all spring stuff sorted out but still an open ended parent hierarchy.

Also, the gradle build can look at jhipster-dependencies as well and we will have all our version specs in one place for both builds!

I'm hoping this won't be a competition between our two approaches, but rather an effort to find the good parts in both!

Tcharl:
@PierreBesson fixed
@deepu105  fixed
@erikkemperman Damn, you were right, this looks far better now!

I just used the kind of result we would have with the @erikkemperman solution on this PR (don't pay attention to the headers, which are now generated by a maven plugin): the result would be customizable by the end-user, and don't add so much overhead.

As we have to release the jhipster-parent and jhipster-dependencies poms on maven central in order to work, I disabled temporarly now unused modules on the other repo (before deleting them if we choose the @erikkemperman solution)

erikkemperman:
@Tcharl Very cool! I was thinking the new modules could be tied with in a multi module setup around the server-lib, so that that also aligns naturally in terms of dependency versions, and all of them can be collectively released with maven-release-plugin.

I guess we can look at the gradle side separately once the maven structure is settled.

Sorry I don't have a lot of time -- in addition to my day job I got involved with a push for Gulp 4.

erikkemperman:
@Tcharl Couple of nitpicking points:

- jhipster uses (4) spaces instead of tabs
- the new poms should maybe not be in a hidden directory (for one thing, Eclipse doesn't see them)

Does that make sense?

Tcharl:
Hi @erikkemperman 

- Arg, you kill me!

- It's visible as long as you configure your project navigator filters enabling "*.resources": I don't think every users need to tweek their parent poms, only advanced ones does (and usually they know how to display them ^^). WDYT? Is it a annoying? I thought adding a new folder on the project hierarchy would be annoying for the UX.

Also, can we continue the discussion on the [generator-jhipster PR](https://github.com/jhipster/generator-jhipster/pull/6075) ?

erikkemperman:
Sorry @Tcharl Im just trying to help things along :-)

erikkemperman:
About the directory, imho it's okay to add `./maven` for this, actually the gradle build also has such a dir for build helpers.

kuzkdmy:
Hi Julien.

kuzkdmy:
I just see that 
QueryService#buildReferringEntitySpecification
is not working with in statements

Thanks.

```
    protected <OTHER, X> Specification<ENTITY> buildReferringEntitySpecification(Filter<X> filter, SingularAttribute<? super ENTITY, OTHER> reference, SingularAttribute<OTHER, X> valueField) {
        if (filter.getEquals() != null) {
            return this.equalsSpecification(reference, valueField, filter.getEquals());
        } else {
            return filter.getSpecified() != null ? this.byFieldSpecified(reference, filter.getSpecified().booleanValue()) : null;
        }
    }
```

jdubois:
I'm closing this in favor of #29

PR #30, opened 2017-10-16
Improve the criteria API

gzsombor:
Setter should return the criteria, so it can be used more concisely : 
Instead of this:
```
BankCriteria bc = new BankCriteria();
bc.setAccountHolderName(new StringFilter());
bc.getAccountHolderName().setContains("john");
```
One can write:
```
BankCriteria bc = new BankCriteria();
bc.setAccountHolderName(new StringFilter().setContains("john"));
```

And fix the missed 'in' fragment handling

### DIFF
--- a/src/main/java/io/github/jhipster/service/QueryService.java
+++ b/src/main/java/io/github/jhipster/service/QueryService.java
@@ -141,6 +141,8 @@ public abstract class QueryService<ENTITY> {
         SingularAttribute<OTHER, X> valueField) {
         if (filter.getEquals() != null) {
             return equalsSpecification(reference, valueField, filter.getEquals());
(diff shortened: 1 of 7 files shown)


DanielFran:
@gzsombor the criteria classes actually implements Serializable but all Filter class did not implement Serializable.

Is this missing the implementation in Filter classes?

gzsombor:
Yes, you've catched me ! I forgot to add it

cbornet:
And SonarQube is very unhappy about that !

gzsombor:
We can make it happy :wink:  https://github.com/jhipster/jhipster/pull/34

DanielFran:
Nice ;) thanks

PR #31, opened 2017-10-29
Add ZonedDateTime converter to Long and vice versa

ramzimaalej:
The goal of the PR is to use String instead of Date because the ZonedDateTime create from the database is slightly different than the original one causing some tests to fail when using Jhipster with mongo and elasticsearch. https://github.com/jhipster/generator-jhipster/pull/6595

### DIFF
--- a/src/main/java/io/github/jhipster/domain/util/JSR310DateConverters.java
+++ b/src/main/java/io/github/jhipster/domain/util/JSR310DateConverters.java
@@ -82,25 +82,25 @@ public final class JSR310DateConverters {
         }
     }
 
-    public static class StringToZonedDateTimeConverter implements Converter<String, ZonedDateTime> {
-        public static final StringToZonedDateTimeConverter INSTANCE = new StringToZonedDateTimeConverter();
+    public static class LongToZonedDateTimeConverter implements Converter<Long, ZonedDateTime> {
+        public static final LongToZonedDateTimeConverter INSTANCE = new LongToZonedDateTimeConverter();
 
-        private StringToZonedDateTimeConverter() {
+        private LongToZonedDateTimeConverter() {
         }
 
-        public ZonedDateTime convert(String source) {
-            return source == null ? null : ZonedDateTime.parse(source);
+        public ZonedDateTime convert(Long source) {
+            return source == null ? null : ZonedDateTime.ofInstant(Instant.ofEpochMilli(source), ZoneOffset.UTC);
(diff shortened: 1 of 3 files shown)


src/main/java/io/github/jhipster/domain/util/JSR310DateConverters.java
```diff
@@ -82,6 +82,28 @@ public ZonedDateTime convert(Date source) {
         }
     }
 
+    public static class StringToZonedDateTimeConverter implements Converter<String, ZonedDateTime> {
+        public static final StringToZonedDateTimeConverter INSTANCE = new StringToZonedDateTimeConverter();
+
+        private StringToZonedDateTimeConverter() {
+        }
+
+        public ZonedDateTime convert(String source) {
```
MathieuAA: Shouldn't it be `public static ZonedDateTime convert(String source)`?

src/main/java/io/github/jhipster/domain/util/JSR310DateConverters.java
```diff
@@ -82,6 +82,28 @@ public ZonedDateTime convert(Date source) {
...
+        }
+    }
+
+    public static class ZonedDateTimeToStringConverter implements Converter<ZonedDateTime, String> {
+        public static final ZonedDateTimeToStringConverter INSTANCE = new ZonedDateTimeToStringConverter();
+
+        private ZonedDateTimeToStringConverter() {
+        }
+
+        public String convert(ZonedDateTime source) {
```
MathieuAA: Shouldn't it be `public static String convert(ZonedDateTime source)`?

src/main/java/io/github/jhipster/domain/util/JSR310DateConverters.java
ramzimaalej: No, it should not.

src/main/java/io/github/jhipster/domain/util/JSR310DateConverters.java
ramzimaalej: No, it should not.

MathieuAA:
@ramzimaalej Oh I understand now!! Yes, that makes sense.
What, however, I don't get is the need to have an instance as there's no state at all in the utils class. Can you enlighten me?

MathieuAA:
Nevermind my comment, I didn't see the whole file was like that...

PR #32, opened 2017-11-08
Unit tests

erikkemperman:
As promised earlier, I have been writing unit tests for the server lib. Unfortunately I am now starting a new project and can probably not spend (a lot of) time on this for the next couple of days / weeks.

So I thought I'd submit what I have, even though it's not entirely complete:

- The coverage for the classes that I have tests for is 100%, but I am still missing QueryService -- which I wasn't really sure how to test to be honest.
- Also not covered are the JSR310 converters, but I was wondering why we don't simply use the ones provided by Spring?

The first commit adds a utility class to assert the contents written to logback. If you like this idea, we could use it in the apps generated by the main generator too.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -47,6 +47,15 @@
         <maven.compiler.source>${java.version}</maven.compiler.source>
         <maven.compiler.target>${java.version}</maven.compiler.target>
 
+        <!-- Sonar properties -->
+        <sonar.jacoco.itReportPath>${project.testresult.directory}/coverage/jacoco/jacoco-it.exec</sonar.jacoco.itReportPath>
+        <sonar.jacoco.reportPath>${project.testresult.directory}/coverage/jacoco/jacoco.exec</sonar.jacoco.reportPath>
+        <sonar.java.codeCoveragePlugin>jacoco</sonar.java.codeCoveragePlugin>
+        <sonar.typescript.lcov.reportPaths>${project.testresult.directory}/coverage/report-lcov/lcov.info</sonar.typescript.lcov.reportPaths>
+        <sonar.sources>${project.basedir}/src/main/</sonar.sources>
+        <sonar.surefire.reportsPath>${project.testresult.directory}/surefire-reports</sonar.surefire.reportsPath>
+        <sonar.tests>${project.basedir}/src/test/</sonar.tests>
+
         <!-- Plugin versions -->
         <maven-compiler-plugin.version>3.6.0</maven-compiler-plugin.version>
         <maven-eclipse-plugin.version>2.10</maven-eclipse-plugin.version>
@@ -56,7 +65,7 @@
         <maven-source-plugin.version>2.4</maven-source-plugin.version>
         <maven-surefire-plugin.version>2.18.1</maven-surefire-plugin.version>
         <nexus-staging-maven-plugin.version>1.6.7</nexus-staging-maven-plugin.version>
-        <sonar-maven-plugin.version>3.2</sonar-maven-plugin.version>
+        <sonar-maven-plugin.version>3.3.0.603</sonar-maven-plugin.version>
         <jacoco-maven-plugin.version>0.7.9</jacoco-maven-plugin.version>
 
         <!-- Dependency versions -->
--- a/pom.xml
+++ b/pom.xml
@@ -51,7 +51,6 @@
         <sonar.jacoco.itReportPath>${project.testresult.directory}/coverage/jacoco/jacoco-it.exec</sonar.jacoco.itReportPath>
         <sonar.jacoco.reportPath>${project.testresult.directory}/coverage/jacoco/jacoco.exec</sonar.jacoco.reportPath>
         <sonar.java.codeCoveragePlugin>jacoco</sonar.java.codeCoveragePlugin>
-        <sonar.typescript.lcov.reportPaths>${project.testresult.directory}/coverage/report-lcov/lcov.info</sonar.typescript.lcov.reportPaths>
         <sonar.sources>${project.basedir}/src/main/</sonar.sources>
         <sonar.surefire.reportsPath>${project.testresult.directory}/surefire-reports</sonar.surefire.reportsPath>
         <sonar.tests>${project.basedir}/src/test/</sonar.tests>
--- a/pom.xml
+++ b/pom.xml
@@ -42,6 +42,7 @@
         <maven.version>3.0.0</maven.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
+        <project.testresult.directory>${project.build.directory}/test-results</project.testresult.directory>
         <maven.build.timestamp.format>yyyyMMddHHmmss</maven.build.timestamp.format>
         <maven.compiler.source>${java.version}</maven.compiler.source>
         <maven.compiler.target>${java.version}</maven.compiler.target>
--- a/src/main/java/io/github/jhipster/async/ExceptionHandlingAsyncTaskExecutor.java
+++ b/src/main/java/io/github/jhipster/async/ExceptionHandlingAsyncTaskExecutor.java
@@ -31,6 +31,8 @@ import org.springframework.core.task.AsyncTaskExecutor;
 public class ExceptionHandlingAsyncTaskExecutor implements AsyncTaskExecutor,
     InitializingBean, DisposableBean {
 
+    static final String EXCEPTION_MESSAGE = "Caught async exception";
+
     private final Logger log = LoggerFactory.getLogger(ExceptionHandlingAsyncTaskExecutor.class);
 
     private final AsyncTaskExecutor executor;
@@ -71,7 +73,7 @@ public class ExceptionHandlingAsyncTaskExecutor implements AsyncTaskExecutor,
     }
 
     protected void handle(Exception e) {
-        log.error("Caught async exception", e);
+        log.error(EXCEPTION_MESSAGE, e);
     }
 
     @Override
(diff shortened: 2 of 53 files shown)


jdubois:
Wow, that's nearly 5,000 lines of code, incredible!! I'll have time to review this at the end of next week.
BTW, you should join the core team at some point, would you be interested? See http://www.jhipster.tech/team/

erikkemperman:
@jdubois Well, I had some free time recently :-) Would be honoured to join the team but have to say upfront I can't really promise I will have a lot of time...

deepu105:
@erikkemperman dont worry as long as you can contribute something once in a while it should be fine :)

erikkemperman:
FYI Rebased this PR to resolve merge conflicts

erikkemperman:
@deepu105 Well, having made that slight reservation, like I said it'd be an honour!

cbornet:
We should push a Sonar analysis to sonarcloud.io !

erikkemperman:
@cbornet I haven’t looked at that, would it be complicated to add?

erikkemperman:
@cbornet 
I managed to get something working on my fork (unittest branch):

https://sonarcloud.io/dashboard?id=io.github.jhipster%3Ajhipster

But I guess someone of the jhipster org should generate a token and we add the `mvn sonar:sonar` command to the end of the Travis build? Is that how this works?

cbornet:
Yes we need to set the token in Travis.
84% is already quite good, thanks for your work ! Now let's get the AAA !

erikkemperman:
Someone with privileges will have to generate a token and add to Travis. AAA would be nice, but maybe let's first review if the tests I wrote make sense :-)

erikkemperman:
Added a commit to include some Sonar properties (copied from generator, mostly).

After this, if I understand things properly, a member of the jhipster org can generate a sonar token and make a commit similar to [this one](https://github.com/erikkemperman/jhipster/commit/031b576cb500b8ddfdb2d5c28dc1601067dcf0bc) except of course with different `sonar.organization` and `sonar.login` properties.

pascalgrimaud:
The token need to be in a environment variable in Travis instead of beeing directly in travis file.

We have some work to do with JHipster organization in SonarCloud, which is currently not used.

src/main/java/io/github/jhipster/async/ExceptionHandlingAsyncTaskExecutor.java
```diff
@@ -71,7 +73,7 @@ private Runnable createWrappedRunnable(final Runnable task) {
     }
 
     protected void handle(Exception e) {
-        log.error("Caught async exception", e);
+        log.error(EXCEPTION_MESSAGE, e);
```
deepu105: I'm not a fan of using constants for log messages unless they are reused in many places. I see that you use them for test assertions as well but testing log messages seems a bit too much to me @jdubois  what do you think?

src/main/java/io/github/jhipster/config/JHipsterConstants.java
```diff
@@ -22,23 +22,20 @@
 /**
  * JHipster constants.
  */
-public final class JHipsterConstants {
+public interface JHipsterConstants {
```
deepu105: I thought this was only adding tests but seems you rewrote some stuff as well so I would like @jdubois to have a look as well before merging this

src/test/java/io/github/jhipster/CoverageTest.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.Set;
+import java.util.TreeSet;
+
+import org.junit.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.reflect.ClassPath;
+
+public class CoverageTest {
```
deepu105: can you add some docs here on what the class does

src/test/java/io/github/jhipster/CoverageTest.java
```diff
@@ -0,0 +1,67 @@
...
+        if (missing.size() > 0) {
+            StringBuffer buffer = new StringBuffer(64 + 32 * missing.size());
+            buffer.append("Missing " + missing.size() + " classes:" + System.lineSeparator());
+            for (String clazz : missing) {
+                buffer.append("  " + clazz + System.lineSeparator());
+            }
+            log.warn(buffer.toString());
+        }
+        // TODO uncomment:
+        // assertThat(missing).isEmpty();
```
deepu105: ??

src/main/java/io/github/jhipster/async/ExceptionHandlingAsyncTaskExecutor.java
erikkemperman: Yes, that’s why I made them constants. Could be made package private if you prefer. Or like you say maybe asserting log messages is too much, but for what it’s worth I actually quite like it. Especially for failure messages, logging is an important function and as such ought to be tested.

src/main/java/io/github/jhipster/config/JHipsterConstants.java
erikkemperman: Just some little tweaks, like those message constants and things like this which changes a class that’s never instantiated to an interface (helps with coverage).

Other than that just minor refactoring so I could spy or override selectively.

No functional changes I think.

src/test/java/io/github/jhipster/CoverageTest.java
erikkemperman: I figured the name of the class would be self-explanatory but maybe not :-)

Actually maybe this test is overly pedantic, it traverses the packages and asserts that every regular class has a `...Test` companion.

I’ll await the feedback here, if this test stays I will add a comment about what it does.

src/test/java/io/github/jhipster/CoverageTest.java
erikkemperman: Once coverage is complete we could uncomment this, it’ll throw a failure when new classes are added but an accompanying test class is missing.

But like I said above, perhaps this is too pedantic.

src/main/java/io/github/jhipster/async/ExceptionHandlingAsyncTaskExecutor.java
erikkemperman: Oh and also, the way the log testing works also keeps the console nice and clean during `mvn test`. So expected exceptions during tests, for example, don’t spew out endless traces.

src/test/java/io/github/jhipster/CoverageTest.java
deepu105: i'm not in favor of doing this ourself, why not us a gradle plugin for that?

erikkemperman: Well, Maven. I’m not too familiar with jacoco but if we can make it fail the build if coverage drops significantly, sure.

But don’t read too much into this particular class, it started as more of a todo list for this PR and only at the end did I add the commented assert as an afterthought

No problem to drop this commit.

deepu105: Then lets remove this. And I don't want builds to fail if coverage drops not yet atleast.

deepu105:
@jdubois i would like to merge this but would like your eyes on it as well. Can you take a quick look

deepu105:
@erikkemperman what is your full name, twitter handle and email  address so that I can start a formal invitation process to the team
If you dont want to share it here you can send it to d4udts@gmail.com

erikkemperman:
FYI Dropped the commit with the coverage test,, and rebased onto latest master.

@deepu105 I will send some details to the address you mentioned, thanks for following up!

pom.xml
```diff
@@ -42,10 +42,20 @@
...
+        <project.testresult.directory>${project.build.directory}/test-results</project.testresult.directory>
         <maven.build.timestamp.format>yyyyMMddHHmmss</maven.build.timestamp.format>
         <maven.compiler.source>${java.version}</maven.compiler.source>
         <maven.compiler.target>${java.version}</maven.compiler.target>
 
+        <!-- Sonar properties -->
+        <sonar.jacoco.itReportPath>${project.testresult.directory}/coverage/jacoco/jacoco-it.exec</sonar.jacoco.itReportPath>
+        <sonar.jacoco.reportPath>${project.testresult.directory}/coverage/jacoco/jacoco.exec</sonar.jacoco.reportPath>
+        <sonar.java.codeCoveragePlugin>jacoco</sonar.java.codeCoveragePlugin>
+        <sonar.typescript.lcov.reportPaths>${project.testresult.directory}/coverage/report-lcov/lcov.info</sonar.typescript.lcov.reportPaths>
```
deepu105: This repo doesn't and will never contain Typescript code so I guess this can be removed

erikkemperman: Ah, good catch, thanks! Sloppy/paste from the generator POM. Will amend now!

src/main/java/io/github/jhipster/async/ExceptionHandlingAsyncTaskExecutor.java
jdubois: I'd rather also have normal Strings as they are more readable, but that's not a huge issue

jdubois:
There are some formatting issues, etc... But this is so big & good I'm merging this, and I'll improve later.

erikkemperman:
Thanks for review and merge! I tried to follow formatting guidelines, so I’ll keep an eye on changes here so hopefully I’ll get it right next time.

About those strings, is it the constants you don’t like or asserting the contents of logs? Because ai wanted to propose doing something similar in generated code...

One benefit aside from checking the logs is that it would avoid cluttering the console — I’ve been getting some errors in Travis about excessive output.

Thoughts?

jdubois:
Let me first finish to polish everything :-)
But yes I'd rather keep the Strings, it's more readable. Let's keep it like that on the server lib, it's less important, and it allows us to test this a bit.

erikkemperman:
All right ;-)

PR #33, opened 2017-11-10
Add a utility to invoke methods reflectively

erikkemperman:
Ref. https://github.com/jhipster/generator-jhipster/pull/6642#issuecomment-343222587

### DIFF
--- a/src/main/java/io/github/jhipster/util/ReflectUtils.java
+++ b/src/main/java/io/github/jhipster/util/ReflectUtils.java
@@ -1,0 +1,134 @@
+/*
+ * Copyright 2016-2017 the original author or authors from the JHipster project.
+ *
+ * This file is part of the JHipster project, see http://www.jhipster.tech/
+ * for more information.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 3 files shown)


src/main/java/io/github/jhipster/util/ReflectUtils.java
```diff
@@ -0,0 +1,134 @@
...
+package io.github.jhipster.util;
+
+import java.lang.reflect.InvocationTargetException;
+
+/**
+ * This utility is intended to be used with optional dependencies that are only used in certain
+ * Spring profiles, such as with Liquibase and Swagger, and potentially stripped entirely from
+ * the classpath. Invoking methods via reflection avoids compile-time errors in those cases.
+ */
+public interface ReflectUtils {
```
deepu105: Btw I think Spring has a built in reflection utils not sure if it will match the requirements but worth a look

erikkemperman: I looked at it briefly but it seemed not to do class loading (and we want to only know about class names) and a bit awkward in handling exceptions. Just my opinion after a cursory glance, mind you.

cbornet: I think it can be used for invoke(...) functions

cbornet:
Does the class loading work fine with devtools hot reload ?

deepu105:
Oh yes making sure hot reload works with this is very important

erikkemperman:
Haven’t tried, to be honest. But the classes/methods that were the original case for this are libraries, not likely hot-patched by users.

jdubois:
I'd rather just refactor the part of the code we need in the generator - here we have a generic classloading library, and I'm afraid it goes too far, as we just want to run H2.
I'm having a look at refactoring the generator and applying the changes to jhipster/jhipster

jdubois:
I'm closing this as I moved the methods from the generator - I only want to have something generic if we need it, otherwise this will be troublesome to maintain as this is highly complex

jdubois:
Yes @erikkemperman please have a look, and then we can still refactor later if we need something generic

PR #37, opened 2017-12-09
Build specification for referring entities field with a join

duderoot:
After the discussion on generator-jhipster issue [6735](https://github.com/jhipster/generator-jhipster/issues/6735) I agree with the proposal from @ctamisier. For that case I add a new method in the QueryService. It will be good if we can cover this new method over a test.
[This ](https://github.com/spring-projects/spring-data-jpa/blob/master/src/test/java/org/springframework/data/jpa/repository/query/SimpleJpaQueryUnitTests.java)Spring Data JPA test looks like a good inspiration  but I did not manage at the moment more than the manual tests 😞 . Nonetheless, this method works for filtering entities in a one-to-one relationship when the referenced filed is part of the owning part.

Fixes #6735

### DIFF
--- a/src/main/java/io/github/jhipster/service/QueryService.java
+++ b/src/main/java/io/github/jhipster/service/QueryService.java
@@ -21,6 +21,7 @@ package io.github.jhipster.service;
 
 import java.util.Collection;
 import javax.persistence.criteria.CriteriaBuilder.In;
+import javax.persistence.criteria.JoinType;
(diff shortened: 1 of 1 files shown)


jdubois:
@gzsombor I don't know this part very well, could you have a look?

duderoot:
I will try in the weekend to create a new PR which is based on the current master.

gzsombor:
A similar patch is merged already, so we can close this.

PR #38, opened 2017-12-14
Just an adjustment for dependencies which should be sorted properly.

jearton:
Just an adjustment for dependencies which should be sorted to compile, optional and test. No dependency deleted or added.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -84,24 +84,22 @@
     </dependencyManagement>
 
     <dependencies>
+        <!--Compile-->
         <dependency>
-            <groupId>com.fasterxml.jackson.datatype</groupId>
-            <artifactId>jackson-datatype-hibernate5</artifactId>
-            <optional>true</optional>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-web</artifactId>
         </dependency>
         <dependency>
-            <groupId>com.fasterxml.jackson.datatype</groupId>
-            <artifactId>jackson-datatype-json-org</artifactId>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-autoconfigure</artifactId>
         </dependency>
         <dependency>
-            <groupId>com.mattbertolini</groupId>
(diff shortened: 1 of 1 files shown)


erikkemperman:
I think re-ordering makes sense, but might give conflicts while upgrading apps. Maybe keep this for the next major release?

jearton:
Yeah, I agree with you.

deepu105:
Erik since this is library why would you think it breaks upgrade? Does this affect generated pom? But I don't see how, anyway im not a maven expert but just curious

erikkemperman:
Sorry! Should not comment before first coffee... Didn't see this was a PR on the server-lib :-)

So forget what I said, it does not apply. Still, might be nice to re-order the deps in the generated POM with the next major.

jdubois:
I just had a quick look: our dependencies are supposed to be sorted alphabetically - then maybe that's not always the case, because sometimes it makes sense to have 2 dependencies close to each other, and also because of the dependency tree resolution (modifying the order might modify the final tree). Ordering them by compile/optional/test could also make sense, but usually people order them alphabetically, at least that's what I usually see.

jearton:
The function of these dependencies is similar to spring-boot-autoconfigure. 

The spring-boot-autoconfigure pom is here: [https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/pom.xml](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/pom.xml)

So I just follow the practice of spring boot.

jearton:
I have never seen this case in other well-known open source projects expect JHipster. Please could you provide some examples which usually seen?

jdubois:
The dependency order has a meaning both in Maven and Gradle. Even worse, they don't use the same algorithm so they might not interpret the same order in the same way.

jdubois:
While I agree it is more readable to have dependencies sorted alphabetically, I'm not merging this as the order of the dependencies is important for Maven.
So I'd rather have close to each other some dependencies that I know will cause trouble.

PR #45, opened 2018-03-12
Add a BaseEntity to extends from all JHipster entity

henri-tremblay:
I've been dreaming of this for a long long time. This is my proposal of a base entity to all entities. I can provide a PR for the generator afterward.


src/main/java/io/github/jhipster/domain/BaseEntity.java
```diff
@@ -0,0 +1,115 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.github.jhipster.domain;
+
+import com.google.common.base.MoreObjects;
```
deepu105: Where does this come from?

src/main/java/io/github/jhipster/domain/BaseEntity.java
```diff
@@ -0,0 +1,115 @@
...
+
+    private static final long serialVersionUID = 1L;
+
+    /**
+     * Unique id for an entity of a given type. Will be {@code null} for an entity that hasn't been saved to its data
+     * storage.
+     */
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
```
deepu105: This would not work for mongo or Cassandra apps as we generate a different Id type

src/main/java/io/github/jhipster/domain/BaseEntity.java
```diff
@@ -0,0 +1,115 @@
...
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+
+        BaseEntity user = (BaseEntity) o;
+        return !(user.getId() == null || getId() == null) && Objects.equals(getId(), user.getId());
```
deepu105: Why user

src/main/java/io/github/jhipster/domain/BaseEntity.java
```diff
@@ -0,0 +1,115 @@
...
+     * <pre>{@code
+     * @Override
+     * protected MoreObjects.ToStringHelper toStringHelper() {
+     *     return super.toStringHelper().add("name", name);
+     * }
+     * }</pre>
+     *
+     * @return {@code ToStringHelper} used to create the final String
+     */
+    protected MoreObjects.ToStringHelper toStringHelper() {
```
deepu105: Do we really need this?

henri-tremblay: No. I found it useful. My personal favorite implementation of the `toString` would be `ReflectionToStringBuilder.toString(this)`. Nothing to override.

src/main/java/io/github/jhipster/domain/BaseEntity.java
henri-tremblay: My bad. Bad cut & paste. Will fix.

src/main/java/io/github/jhipster/domain/BaseEntity.java
henri-tremblay: Yes... I was pretty sure something like this would happen. Choices are:
* Do not use a base class for mongo and cassandra
* Use a `BaseJpaEntity` on top of `BaseEntity`
* Use a `BaseJpaEntity` and a `BaseNoqlEntity` or something alike
* Put the `BaseEntity` in the generator instead of here to adapt

src/main/java/io/github/jhipster/domain/BaseEntity.java
henri-tremblay: That's for the `toString`. See comment below.

henri-tremblay:
I'm not a big fan of the JHipster jar. I agree that it removes a lot of code from the generation and is much more joyful to maintain but it still makes things less flexible. For instance, as soon as you want to have a property in the `Cache` object, you need to add your own `Cache` in the `ApplicationProperties`. So putting `BaseEntity` directly in the generated code is fine with me.

I don't like to see the tons of cut&paste for `equals`, `hashCode` and `toString`. Also, a link between all these entities is sometimes useful. BTW, my next step is to build a `BaseResourceIntTest`.

Here is an alternative that you might find nicer. It solves the problem with cassandra and mongo as well I think (unless the id is not a Long. In that case we will need a generic). We can add a BaseJpaEntity on top of it.

```
/**
 * Base class of all entities generated by JHipster.
 */
public abstract class BaseEntity implements Serializable {

    private static final long serialVersionUID = 1L;

    public BaseEntity() {}

    public abstract Long getId();

    /**
     * Equality is based on the id. Two instances of the exact same class and sharing the same id are considered equals.
     * If the ids are {@code null}, two instances will be equals if they actually are the same instance.
     *
     * @param o Object to compare
     * @return if the two instances are equals
     */
    @Override
    public boolean equals(Object o) {
        if (this == o) {
            return true;
        }
        if (o == null || getClass() != o.getClass()) {
            return false;
        }

        BaseEntity entity = (BaseEntity) o;
        return !(entity.getId() == null || getId() == null) && Objects.equals(getId(), entity.getId());
    }

    /**
     * Hashcode based on the entity id.
     *
     * @return the id hashcode or 0 if the id is null
     */
    @Override
    public int hashCode() {
        return Objects.hashCode(getId());
    }

    @Override
    public String toString() {
        return ReflectionToStringBuilder.toString(this);
    }
}
```

src/main/java/io/github/jhipster/domain/BaseEntity.java
deepu105: Yes I got that, my question was do we have a dependency already for this? i dont remember using something from google thats why?

deepu105:
If we are doing this, for what ever reason, I guess it should be generated and not in the jar so people could easily customize it, say for example add audit fields, this is what I do with my entity audit module, I create a base auditEntity and extend every entity with that

src/main/java/io/github/jhipster/domain/BaseEntity.java
henri-tremblay: It's IntelliJ's fault :-) It uses it to generate the `toString`. And I did it so much on my project that I thought it was coming from JHipster.

henri-tremblay:
Yes. My `AbstractAuditingEntity` now extends `BaseEntity`. So, we should move the class to the generator?

jdubois:
Yes a solution to my second issue/question is that we generate this in the application.

For the equals/hashcode I'm reluctant in using reflection or a library to do this:

- people should re-code this to have a correct equals/hashcode
- I've had some very bad performance issues with the Apache Commons equals/hashcode implementation, as this can be heavily used when using Collections, and doing reflection on each field is slow at scale

henri-tremblay:
@jdubois I don't get it. The equals and hashCode are the same that was is currently generated. Only the `toString` is generated by reflection. But I don't expect `toString` to be a performance bottleneck.

jdubois:
Oh my fault @henri-tremblay - I saw the toString and I thought (wrongly) it was the same for equals/hashcode
Indeed the toString should not be an issue

henri-tremblay:
But fine. Let me do a PR directly on the generator. I will close this one when it's done

Blackdread:
If I do not mistaken BaseEntity would not work with Hibernate @Immutable
See [doc](http://docs.jboss.org/hibernate/orm/5.2/userguide/html_single/Hibernate_User_Guide.html#entity-immutability) and [javadocs](https://docs.jboss.org/hibernate/orm/5.2/javadocs/org/hibernate/annotations/Immutable.html).

The doc is not very clear about root entity but javadoc specifies it.
> @Immutable must be used on root entities only

DanielFran:
@jdubois @deepu105 Are we moving forward with that implementation?
From the last comments this should go directly to the generator anyway...

DanielFran:
@pascalgrimaud Closing this one, this has been open for so long...
@henri-tremblay If needed we can open it in the generator.

jdubois:
@DanielFran yes I'm pretty sure people will complain about the "complexity" of JHipster with this: it adds another class, and also require that people understand how this works.

henri-tremblay:
Sad sad. I think it was removing a lot of code duplication. Right now, it's the first thing I do after generating something. I create a base class. The reason I never finished was that is was complicated to get something that works nicely out of RDBMS. Especially because I know those other implementations less.

PR #46, opened 2018-03-13
Update Spring Platform to Cairo-RC1, dependencies and plugins versions

DanielFran:
See https://spring.io/blog/2018/03/12/spring-io-platform-cairo-rc1

### DIFF
--- a/jhipster-dependencies/pom.xml
+++ b/jhipster-dependencies/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <groupId>io.spring.platform</groupId>
         <artifactId>platform-bom</artifactId>
-        <version>Cairo-BUILD-SNAPSHOT</version>
+        <version>Cairo-RC1</version>
         <relativePath />
     </parent>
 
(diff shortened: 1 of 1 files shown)


DanielFran:
@jdubois , with this commit you could create beta version of v5?

jdubois:
@DanielFran I'm hoping to do a beta release on Thursday. We still have a few issues with Webpack (should be solved) and OAuth2 (probably won't be solved)

erikkemperman:
@DanielFran Argh we seem to have been doing similar things yesterday, although I did not submit my [PR](https://github.com/jhipster/jhipster/pull/47) until this morning because of an issue with infinispan-spring-boot-starter. I notice a couple of other little differences, maybe we should compare?

- You did not update infinispan-boot-starter, perhaps because of the issue I ran into?
- Some of the version properties you updated are now at the same value as spring-boot-dependencies, do you agree it would be better to remove those and have alignment automatically going forward?
- I have some updates that you don't, and vice versa...

DanielFran:
@erikkemperman we did the same!
- I did not update infinispan-boot-starter because I did not know that exist a new dependency
- I updated everything according to spring boot dependencies but forget to remove the declaration of those dependencies
- Seems that I missed mssql-jdbc update

Should I proceed with the PR and just remove the declaration of some dependencies?

Probably we need to take care of Infinispan in another PR?

erikkemperman:
@DanielFran I'm not sure what to do with infinispan -- on the one hand it's related to various other artifacts that are being updated (the infinispan-spring-starter that we use now is for spring boot 1.x it seems), but on the other hand it is causing problems that I don't know how to solve...

But yes, maybe we should try to do that separately.

I was just seeing if I could easily merge our efforts by rebasing my branch onto yours, will let you know shortly how that works out all right?

erikkemperman:
@DanielFran All right, I think I managed to merge our efforts:
https://github.com/erikkemperman/jhipster/commits/upgrade-deps-rebase

It's now being tested:
https://travis-ci.org/erikkemperman/jhipster/builds/353397119?utm_source=github_status&utm_medium=notification

This excludes the infinispan stuff (need to do that before v5 though, in my opinion).

If this build passes, is it allright with you if I make a new combined PR and close the two separate ones? Obviously only if you agree with the result :-)

@jdubois As a matter of preferred style, I was just wondering whether it is better to make many small commits (one per upgraded dependency, for example) or squash these all in one topical commit?

DanielFran:
@erikkemperman yes, let's do a combined PR! ;)

erikkemperman:
All right, I will push it tonight! Thanks for the quick feedback!

erikkemperman:
Closing this in favour of the combined effort: https://github.com/jhipster/jhipster/pull/48

PR #47, opened 2018-03-14
Upgrade dependencies

erikkemperman:
(Don't merge yet, problem with infinispan).

I've made most dependencies up-to-date, and where possible aligned with spring-boot-dependencies (makes the POM significantly smaller). Also we're now using spring-platform Cairo.RC1 so we don't have to rely on spring's snapshot repository.

There is one failing Travis build, unfortunately, and I am not sure how to fix it. See
https://travis-ci.org/erikkemperman/jhipster/builds/352921745?utm_source=github_status&utm_medium=notification
I've updated a bit of generator code to accomodate the infinispan upgrade, but I honestly don't think that's where the issue comes from:
https://github.com/erikkemperman/generator-jhipster/commit/71e075df64acd1268e31539bf5e2ae68f829d58f

It looks like the way infinispan is used in the generator, specifically in `config/CacheConfiguration.java`, is different from what the README in infinispan-spring-boot-starter says, because it avoids being managed by Spring for some reason.

Unfortunately I don't really have the expertise to deal with this adequately, nor the time to attain it just now... However I didn't want to let the other changes go to waste and therefore thought I might submit this anyway. If anyone knowlegable about the JHipster's usage of infinispan could chime in here, that would be greatly appreciated!

The most significant bump here is probably logstash-logback-encoder, to 5.0. From the changelog I don't expect trouble, but since I usually use log4j myself perhaps someone else should take a look to make sure.

The last commit should not be merged, it's just changing the refs where Travis should get the generator.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,8 +32,8 @@ env:
     - SPRING_JPA_SHOW_SQL=false
 
     # if JHIPSTER_BRANCH value is release, use the release from NPM
-    - JHIPSTER_REPO=https://github.com/jhipster/generator-jhipster.git
-    - JHIPSTER_BRANCH=master
+    - JHIPSTER_REPO=https://github.com/erikkemperman/generator-jhipster.git
+    - JHIPSTER_BRANCH=single-repo
 
   matrix:
     - JHIPSTER=ngx-default PROFILE=prod PROTRACTOR=1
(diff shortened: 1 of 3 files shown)


erikkemperman:
Closing this in favour of the combined effort: https://github.com/jhipster/jhipster/pull/48

PR #48, opened 2018-03-14
Upgrade dependencies (merged #46 and #47)

erikkemperman:
This is a merge of two "competing" (complementary) PRs:
https://github.com/jhipster/jhipster/pull/46
https://github.com/jhipster/jhipster/pull/47

I will close those in favour of this one.

@DanielFran  I hope this is satisfactory to you, if not please let me know!
@jdubois  If you prefer I can squash all my commits into one.

### DIFF
--- a/jhipster-dependencies/pom.xml
+++ b/jhipster-dependencies/pom.xml
@@ -72,7 +72,7 @@
         <lz4-java.version>1.4.1</lz4-java.version>
         <metrics-spring.version>3.1.3</metrics-spring.version>
         <mongobee.version>0.13</mongobee.version>
-        <mssql-jdbc.version>6.2.2.jre8</mssql-jdbc.version>
+        <mssql-jdbc.version>6.4.0.jre8</mssql-jdbc.version>
         <oracle-jdbc.version>12.2.0.1</oracle-jdbc.version>
         <problem-spring-web.version>0.22.2</problem-spring-web.version>
         <prometheus-simpleclient.version>0.3.0</prometheus-simpleclient.version>
--- a/jhipster-dependencies/pom.xml
+++ b/jhipster-dependencies/pom.xml
@@ -59,7 +59,6 @@
         <couchmove.version>1.0.1</couchmove.version>
         <couchbase-testcontainer.version>1.11</couchbase-testcontainer.version>
         <cucumber.version>1.2.5</cucumber.version>
-        <dropwizard-metrics.version>3.2.6</dropwizard-metrics.version>
         <guava.version>24.0-jre</guava.version>
         <jhipster-server.version>2.0.0-20180313.135500-1</jhipster-server.version>
         <hazelcast.version>3.9.3</hazelcast.version>
@@ -291,27 +290,7 @@
             </dependency>
             <dependency>
                 <groupId>io.dropwizard.metrics</groupId>
-                <artifactId>metrics-annotation</artifactId>
-                <version>${dropwizard-metrics.version}</version>
-            </dependency>
-            <dependency>
-                <groupId>io.dropwizard.metrics</groupId>
                 <artifactId>metrics-jcache</artifactId>
-                <version>${dropwizard-metrics.version}</version>
-            </dependency>
-            <dependency>
-                <groupId>io.dropwizard.metrics</groupId>
-                <artifactId>metrics-json</artifactId>
-                <version>${dropwizard-metrics.version}</version>
-            </dependency>
-            <dependency>
-                <groupId>io.dropwizard.metrics</groupId>
-                <artifactId>metrics-jvm</artifactId>
-                <version>${dropwizard-metrics.version}</version>
-            </dependency>
-            <dependency>
-                <groupId>io.dropwizard.metrics</groupId>
-                <artifactId>metrics-servlet</artifactId>
                 <version>${dropwizard-metrics.version}</version>
             </dependency>
             <dependency>
@@ -334,11 +313,6 @@
                 <groupId>com.github.differentway</groupId>
                 <artifactId>couchmove</artifactId>
                 <version>${couchmove.version}</version>
-            </dependency>
-            <dependency>
-                <groupId>com.couchbase.client</groupId>
-                <artifactId>java-client</artifactId>
-                <version>${couchbase-client.version}</version>
             </dependency>
             <dependency>
                 <groupId>com.microsoft.sqlserver</groupId>
--- a/jhipster-dependencies/pom.xml
+++ b/jhipster-dependencies/pom.xml
@@ -53,7 +53,6 @@
         <maven.compiler.target>${java.version}</maven.compiler.target>
 
         <!-- Dependency versions -->
-        <assertj.version>3.9.1</assertj.version>
         <aws-java-sdk-ssm.version>1.11.293</aws-java-sdk-ssm.version>
         <bucket4j.version>3.1.1</bucket4j.version>
         <cassandra-driver.version>3.4.0</cassandra-driver.version>
--- a/jhipster-dependencies/pom.xml
+++ b/jhipster-dependencies/pom.xml
@@ -58,7 +58,6 @@
         <cassandra-unit-spring.version>3.3.0.2</cassandra-unit-spring.version>
         <couchmove.version>1.0.1</couchmove.version>
         <couchbase-testcontainer.version>1.11</couchbase-testcontainer.version>
-        <couchbase-client.version>2.5.5</couchbase-client.version>
         <cucumber.version>1.2.5</cucumber.version>
         <dropwizard-metrics.version>3.2.6</dropwizard-metrics.version>
         <guava.version>24.0-jre</guava.version>
--- a/jhipster-dependencies/pom.xml
+++ b/jhipster-dependencies/pom.xml
@@ -61,7 +61,6 @@
         <cucumber.version>1.2.5</cucumber.version>
         <guava.version>24.0-jre</guava.version>
         <jhipster-server.version>2.0.0-20180313.135500-1</jhipster-server.version>
-        <hazelcast-hibernate52.version>1.2.3</hazelcast-hibernate52.version>
         <hazelcast-wm.version>3.8.3</hazelcast-wm.version>
         <hikaricp.version>2.7.8</hikaricp.version>
         <infinispan.version>9.1.6.Final</infinispan.version>
@@ -200,11 +199,6 @@
                 <groupId>info.cukes</groupId>
                 <artifactId>cucumber-spring</artifactId>
                 <version>${cucumber.version}</version>
-            </dependency>
-            <dependency>
-                <groupId>com.hazelcast</groupId>
-                <artifactId>hazelcast-hibernate52</artifactId>
-                <version>${hazelcast-hibernate52.version}</version>
             </dependency>
             <dependency>
                 <groupId>com.hazelcast</groupId>
--- a/jhipster-dependencies/pom.xml
+++ b/jhipster-dependencies/pom.xml
@@ -62,7 +62,6 @@
         <guava.version>24.0-jre</guava.version>
         <jhipster-server.version>2.0.0-20180313.135500-1</jhipster-server.version>
         <hazelcast-wm.version>3.8.3</hazelcast-wm.version>
-        <hikaricp.version>2.7.8</hikaricp.version>
         <infinispan.version>9.1.6.Final</infinispan.version>
         <infinispan-cloud.version>9.1.6.Final</infinispan-cloud.version>
         <infinispan-spring-boot-starter.version>1.0.0.Final</infinispan-spring-boot-starter.version>
--- a/jhipster-dependencies/pom.xml
+++ b/jhipster-dependencies/pom.xml
@@ -101,7 +101,6 @@
         <!-- Plugin versions -->
         <maven-eclipse-plugin.version>2.10</maven-eclipse-plugin.version>
         <maven-gpg-plugin.version>1.6</maven-gpg-plugin.version>
-        <maven-javadoc-plugin.version>3.0.0</maven-javadoc-plugin.version>
         <maven-source-plugin.version>3.0.1</maven-source-plugin.version>
         <maven-surefire-plugin.version>2.21.0</maven-surefire-plugin.version>
         <nexus-staging-maven-plugin.version>1.6.8</nexus-staging-maven-plugin.version>
--- a/jhipster-dependencies/pom.xml
+++ b/jhipster-dependencies/pom.xml
@@ -101,7 +101,6 @@
         <!-- Plugin versions -->
         <maven-eclipse-plugin.version>2.10</maven-eclipse-plugin.version>
         <maven-gpg-plugin.version>1.6</maven-gpg-plugin.version>
-        <maven-surefire-plugin.version>2.21.0</maven-surefire-plugin.version>
         <nexus-staging-maven-plugin.version>1.6.8</nexus-staging-maven-plugin.version>
         <sonar-maven-plugin.version>3.4.0.905</sonar-maven-plugin.version>
         <jacoco-maven-plugin.version>0.8.0</jacoco-maven-plugin.version>
--- a/jhipster-dependencies/pom.xml
+++ b/jhipster-dependencies/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <groupId>io.spring.platform</groupId>
         <artifactId>platform-bom</artifactId>
-        <version>Cairo-BUILD-SNAPSHOT</version>
+        <version>Cairo-RC1</version>
         <relativePath />
     </parent>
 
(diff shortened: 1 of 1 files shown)


erikkemperman:
Still todo: upgrades are available for infinispan-spring-starter, but it needs work. Even the updates are already a bit stale, I've asked about possibly an update: https://github.com/infinispan/infinispan-spring-boot/issues/44

erikkemperman:
Hmm, I had to restart two Travis builds (flaky webdriver?) but now it's all green :-)

jdubois:
Don't squash your commits - it's easier to rollback one dependency if we see an issue

erikkemperman:
@jdubois All right, thanks, I’ll leave them split up. Any suggestions who might be able to come to the rescue with the infinispan-spring-starter problem? I could make a (failing) PR to demonstrate the issue if that helps?

jdubois:
@erikkemperman for Infinispan you can ask @srinivasa-vasu -> he works for Red Hat and did our Infinispan support, so he's probably the one who this best!!
Otherwise, can't we stay on an older version? There's no need to have the latest bleeding edge version all the time

erikkemperman:
That’s certainly true, but we’re now using a starter built for Spring-Boot 1.x, and we probably need some changes in the generator to accommodate the newer version which would be nice to have prior to JH v5, it seems to me?

By the way, it just occurred to me that the Travis setup here might be wrong: aren’t we testing the generator against the previously deployed snapshot instead of the PR at hand?

I guess you have good reason to stick with JFrog, instead of deploying snapshots to Sonatype?

jdubois:
- I didn't know the starter is for Spring Boot 1 -> if there is no way to have Infinispan with Spring Boot 2 right now, we could disable the option temporarily, what do you think? We're not going to delay JHipster 5 for this reason
- For Travis, I'm not sure of your issue, but for me we should test against the current PR or branch at the moment. We'll use a deployed release when we are more stable.
- I'm sticking to JFrog because I don't have the permissions to deploy snapshots to Sonatype, from my understanding this is now forbidden (at least for mere mortals like me). Anyway we also use JFrog for JHipster IDE, for playing with NuGet/Chocolatey, etc... And we have a complete control of our repo there, which is much better for testing.

jdubois:
@erikkemperman I'm merging this but we need a strategy for Infinispan - can you explain the issue? @srinivasa-vasu would you have time to help?
Otherwise we can also put a big warning on the option - we are still in BETA mode, that's not a problem

erikkemperman:
I’m not saying infinispan-starter at the current version won’t work, to be clear, just that it’s POM lists spring-boot stuff at 1.5...

About snapshots, I did not know sonatype changed policy, but it sounds like jfrog would be preferred anyway.

But then we should have the artifact versions of jhipster-server and jhipster-dependencies in github at the exact same values as the one listed in the generator — otherwise it won’t use the build that Travis just made from the PR with `mvn install` but will instead just fetch the remote snapshot!

jdubois:
By the way maybe we can now remove the Spring snapshot repository?? We're on milestones/RC releases everywhere, am I correct?
I'm not sure of your last issue - I'm now deploying the BOM to JFrog, and will use from generator-jhipster, so that should test it

erikkemperman:
Sorry, I'm not really explaining myself properly. Thanks for review and merge, I will make a new issue with the Travis issue (if that's what it is) as it doesn't really belong here.

deepu105:
@jdubois btw @srinivasa-vasu works for Pivotal now :)

jdubois:
Oh congrats @srinivasa-vasu !

erikkemperman:
Huh, we have issues disabled on this repo?

OK I'll try to make my point here then:

- Travis clones the repository and does `mvn install` to its local `~/.m2/repository`. It will therefore use the versions numbers listed in the POMs. These are both still at 2.0.0-SNAPSHOT
https://github.com/jhipster/jhipster/blob/master/jhipster-server/pom.xml#L14
https://github.com/jhipster/jhipster/blob/master/jhipster-dependencies/pom.xml#L14

So the build with the changes from the PR at hand will be in `~/.m2/repository/io/github/jhipster/.../2.0.0-SNAPSHOT/...`.
- Then Travis clones the generator, with specified url and branch, and run those builds. But those, because the generator refers to a different version of the BOM, will *not* be built against the PR code but against the previously deployed BOM:
https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/pom.xml.ejs#L114
https://github.com/jhipster/generator-jhipster/blob/master/generators/server/templates/gradle.properties.ejs#L28
- Furthermore, the jhipster-server module will not be compiled against the PR version of jhipster-dependencies but also against an earlier snapshot, because of
https://github.com/jhipster/jhipster/blob/master/jhipster-server/pom.xml#L8 (which is again different from the version listed in the generator).
- And this is yet again different from the jhipster-server version that is dependency-managed by jhipster-dependencies (and used by the generator for that reason):
https://github.com/jhipster/jhipster/blob/master/jhipster-dependencies/pom.xml#L63

This is all because JFrog assigns unique versions for snapshots, if I understand the issue properly, which IMHO goes against the whole point of deploying snapshots in the first place...

I'm not sure if, with this limitation, we can things 100% lined up, but probably the best we could do is:

- deploy the jhipster-dependencies, say we get back version X from JFrog.
- change artifact version of jhipster-dependencies in github to X so Travis builds for future PRs will install it as X locally
- change the parent reference in jhipster-server to X
- deploy jhipster-server to JFrog, say we get back version Y
- change the artifact version of jhipster-server in github to Y so Travis builds for future PRs will install it as Y locally
- change the jhipster-server property in jhipster-dependencies BOM to Y so future local builds will be mutually consistent.
- change the jhispter-dependencies property in the generator maven and gradle setup to X so future generator builds will use the latest deployed snapshots, and future jhipster builds will use the locally built PR artifacts.

This still leaves one quirk which I don't think we can avoid with JFrog then, which is that the deployed version of jhipster-dependencies will have a version property for jhispter-server at Y-1 instead of Y.

Sorry for being so verbose, it's all a bit confusing and I wanted to be explicit about the problem I'm seeing.

None of this kind of thing is happening with another multi-module maven project I am involved in, which uses Sonatype for snapshots, but I have to admit we haven't deployed anything new for a while so perhaps their policy did change in the mean time. Perhaps this is why Spring have their own snapshot repository.

(Yes we could remove that snapshot repo reference now, by the way, because we are using Cairo.RC1).

erikkemperman:
... And one more thing (sorry) -- none of this helps with the case where we need mutually dependent changes in jhispter-server/-dependencies and generator :-(

jdubois:
Oh yes @erikkemperman you are totally correct. Let me sum up:

- For the issues -> I disabled them here as we had people opening lots of issues which were in fact for jhipster/generator-jhipster so I thought it would be better to force them to go there. I agree it's bad for us, but if users could just read the documentation before opening an issue, we would have JHipster 5 out since 2 months :-)
- For JFrog -> yes you totally see my issue, and I'm currently doing exactly that... That's very annoying, but I hope it's temporary. I will ask my friends at JFrog about that.

jdubois:
Oh that's an option in JFrog!!!!!!!!!!
Let me try that -> how about putting both repos on "2.0.0-SNAPSHOT",  think that's what we we all want, is that OK @erikkemperman ?

erikkemperman:
Oh great, if that’s an option I totally think we should!

So put everything everywhere (all the lines I linked earlier) at 2.0.0-SNAPSHOT, artifact versions and parent versions and properties, also in the generator, and then the whole issue would go away!

At that point we’re still not sure if the just-merged PR is good (it wasn’t actually tested in the generator because of all this).

And for releases, we’d just need to change the jhipster-server.version property in the BOM to the new release value prior to deploying the BOM and everything should line up!

jdubois:
Yes that's what I was dreaming of! I have no idea what JFrog has this other "strange" option by default! I just needed to look around the admin panel.
Let's test this!

erikkemperman:
Cool! Let me know how that goes? I’ll be out for a couple of hours at the hospital with my GF, but will check back this afternoon.

jdubois:
Oh, hope everything fine??? GF is more important than a Maven build, you know :-)

It all looks very good:
- I'm doing some last tests and will commit
- We can then have a look at how we can improve everything, but that should already be far better

erikkemperman:
Thanks, yes I have my priorities in that order :-)

pascalgrimaud:
@erikkemperman @jdubois : I did some changes few weeks ago, for different cloning in Travis builds. But with the recent merge, maybe there are some work to do.
Let me know if you need my help on this part, because you go too fast and I didn't follow you closely

jdubois:
So it looks quite good but we have 2 issues:

- Indeed Infinispan fails... Maybe we should exclude the option, we have no time for this. @erikkemperman did you have a closer look?
- I am not sure at all if Travis is OK or not. I don't think it changed that much, and you're the boss here @pascalgrimaud so your help is welcome. BTW we should have more React builds now.

erikkemperman:
About Infinispan, there are a few package changes in the new version, so generator code has to be changed. But even with those, it compiles but fails to run.

I can link to the commit I made when I get home.

I’ve requested a new release for infinispan-spring-starter, but no response yet. I’ll link that issue as well this afternoon.

jdubois:
OK @erikkemperman I'll let you decide, you seem to know that very well
But the build is broken so we should do something today

erikkemperman:
Well, thanks but not really — I couldn’t figure it out properly! But I can make a first attempt and call in for help at that point.

When you say the build is broken, you mean the Travis/snapshot mess? Or something else?

jdubois:
The Travis build with Infinispan is broken, only that. And it was working before, can't we just rollback?

erikkemperman:
Not by reverting, that change was in DanielfFran’s commit. But you can change it back easily:
https://github.com/jhipster/jhipster/pull/48/files#diff-78a0a62b9524ac5cc8a07f55a4a29683L73
And the next line.

jdubois:
I didn't do anything and now the JFrog repo is totally broken :-(
I am totally lost :-((

jdubois:
Sorry everyone, I'm going back to the old system, where SNAPSHOTs did not really work, and which was a pain.
I have no idea what happened, but using "real SNAPSHOTs" does not seem to work very well.

erikkemperman:
Uhm no idea what happened? What broke exactly? With sonatype snapshots there was sometimes an issue with re-deploying — if I remember correctly it was due to caching and a short wait would help (not sure if there was any way to invalidate the cache without intervention from support folks).

If you go back to those “timestamp snapshots”, could you also change the artifact versions themselves? That way the Travis build might be at least partially informative!

jdubois:
OK so reverting to the "old" Infinispan works!!! Let me rollback this, and hopefully we can have a stable build back

jdubois:
@erikkemperman what broke is that the "SNAPSHOT" were not available anymore :-(
So we're back with the timestamps. They are annoying but at least they work!

erikkemperman:
@jdubois Well, that sounds like a pretty bad bug on their part; should we report it? If it were fixed the whole process would be so much easier!

jdubois:
Yes they are very good friends and I will get back to them. At least, I think the product isn't working by default as we would expect.

pascalgrimaud:
@erikkemperman : can you describe the workflow you use to test locally the change in this project ?

My workflow:
- change the version in all pom.xml to a non existing one (`0.0.0-TRAVIS`)
- launch: `./mvnw clean install -Dgpg.skip=true`
- generate a new project with generator-jhipster, using master branch
- change the pom.xml of the generated project, and put `0.0.0-TRAVIS`
- launch all tests (back, front and protractor)

Am I right ? If yes, I can start working on these changes for our current build here, in this project -> jhipster/jhipster

erikkemperman:
@pascalgrimaud Well I don’t usually bother with a non-existent version but that’d be cleaner, sure. Otherwise yes, make sure all the versions line up and check that I’m only testing locally built artifacts.

I have to change five in jhipster/jhipster before I do mvn install:
- artifact version of jhipster-parent (top-level POM, the name is confusing it’s not anyone’s parent actually)
- artifact version of jhipster-dependencies
- artifact version of jhipster-server
- parent version of jhipster-server
- jhipster-server.version property in jhipster-dependencies

And then two properties in the generator, jhipster-dependencies.version for Maven and Gradle. I usually do this once in the generator prior to generating projects so I don’t have to edit them all (and `yarn link` of course).

Are you thinking of making Travis do something similar, with sed or something like that?

pascalgrimaud:
Thanks for describing your workflow Eric.

Yes, I was thinking to use some `sed` to make sure the version doesn't officially exist, for our Travis builds. 
I'm starting to work on this.

erikkemperman:
That would be great! If I may give a suggestion, maybe put 0.0.0-Travis in the github repos of these modules in all the places I mentioned, otherwise you’re going to have to do tricky multi line magic with sed I guess.

pascalgrimaud:
I don't think it's a good idea.
1) put 0.0.0-TRAVIS everywhere in these modules will imply:
- Julien will have to change the version before release
- Commit -> and the build can fail
- Do the release
- Put back to 0.0.0-TRAVIS
- Commit again

2) I prefere keep as currently
- Travis changes to 0.0.0-TRAVIS
- install in his local ~/.m2/ these special libs / jar / bom
- generate a new JHipster project and change to use 0.0.0-TRAVIS
- do all tests
- kill the VM / Container :)

erikkemperman:
Well, changing those versions before release has to be done anyway, but all right :-)

Are you planning to test all the generator’s Travis samples?

I’m curious what your sed expressions will be, I’m no guru but always have a hard time with multi line matches for some reason.

pascalgrimaud:
@erikkemperman : here my code

```bash
#!/bin/bash

# artifact version of jhipster-parent (top-level POM, the name is confusing it’s not anyone’s parent actually)
sed -i '1,10 s/<version>.*<\/version>/<version>0.0.0-TRAVIS<\/version>/1' ./pom.xml

# artifact version of jhipster-dependencies
sed -i '11,20 s/<version>.*<\/version>/<version>0.0.0-TRAVIS<\/version>/1' ./jhipster-dependencies/pom.xml

# artifact version of jhipster-server
# parent version of jhipster-server
sed -i '1,20 s/<version>.*<\/version>/<version>0.0.0-TRAVIS<\/version>/1' ./jhipster-server/pom.xml

# jhipster-server.version property in jhipster-dependencies
sed -i 's/<jhipster-server.version>.*<\/jhipster-server.version>/<version>0.0.0-TRAVIS<\/version>/1' ./jhipster-dependencies/pom.xml
```

Not perfect but it seems to work. Time to go at work now, I will work back on it later :-)

erikkemperman:
@pascalgrimaud Ah, that's cute, I did not know it was possible to limit on line numbers this way! Bit dangerous if someone inserts a big comment at the top or something like that, but probably that's not a big deal.

pascalgrimaud:
Not a big deal, because his build will failed :laughing:

erikkemperman:
Yeah, but imagine the surprise of having your build fail when all you did is add a comment! The reason why it failed will be *hard* to track down.

If I find time I might try the multi-line match approach, but this is definitely already really nice!

erikkemperman:
@pascalgrimaud One little error in your script, I think -- you will be replacing
`<jhipster-server.version>...</jhipster-server.version>`
with
`<version>...</version>`

Just for fun and learning I've played around with multiline approach, I think something like this would also work, and doesn't use line numbers:

```sh
#!/bin/bash

if [[ $VERSION == '' ]]
then
  VERSION=0.0.0-TRAVIS
fi

# artifact version of jhipster-parent
sed -i '/<artifactId>jhipster-parent<\/artifactId>/{$!{N;s/<version>.*<\/version>/<version>'$VERSION'<\/version>/1}}' pom.xml

# artifact version of jhipster-dependencies
sed -i '/<artifactId>jhipster-dependencies<\/artifactId>/{$!{N;s/<version>.*<\/version>/<version>'$VERSION'<\/version>/1}}' jhipster-dependencies/pom.xml

# jhipster-server.version property in jhipster-dependencies
sed -i 's/<jhipster-server.version>.*<\/jhipster-server.version>/<jhipster-server.version>'$VERSION'<\/jhipster-server.version>/1' ./jhipster-dependencies/pom.xml

# parent version of jhipster-server
sed -i '/<artifactId>jhipster-dependencies<\/artifactId>/{$!{N;s/<version>.*<\/version>/<version>'$VERSION'<\/version>/1}}' jhipster-server/pom.xml

# artifact version of jhipster-server
sed -i '/<artifactId>jhipster<\/artifactId>/{$!{N;s/<version>.*<\/version>/<version>'$VERSION'<\/version>/1}}' jhipster-server/pom.xml
```

Using a variable for the version here, the same script could be used for releases:

```sh
VERSION=2.0.0-RELEASE ./travis/scripts/bump.sh
# do the release
VERSION=2.1.0-SNAPSHOT ./travis/scripts/bump.sh
# set for next dev cycle
```

pascalgrimaud:
yes, it's better! I just coded it, this morning during 20min in the train. No time to clean code :)

jdubois:
Guys, big news, it looks like uploading snapshots to the Sonatype OSS repo is possible: https://maven.apache.org/guides/mini/guide-central-repository-upload.html

This is a game changer for us.
I'm trying to migrate there.

mraible:
It’s been possible for quite some time. I used to do it for AppFuse 10+ years ago! ;)

PR #44, opened 2018-03-09
Fix range filtering on referring entity

Blackdread:
Hi,I propose these changes to QueryService to solve issue to have range filtering for referring entity.
Pull request relates to issue [7195](https://github.com/jhipster/generator-jhipster/issues/7195).

For range filtering methods, I used **buildReferringEntitySpecification**, do you prefer **buildRangeReferringEntitySpecification** ? So it does not add this feature by default so user have to manually edit generated filter service classes to use the range filtering?

I used **root.get(reference)** for SingularAttribute but **root.join(reference)** for SetAttribute

@jdubois @gzsombor @DanielFran

PR #49, opened 2018-03-15
Fix range filtering on referring entity (fix commit history)

Blackdread:
Hi,
(Maybe can close other pull request #44?)

I propose these changes to QueryService to solve issue to have range filtering for referring entity.
Pull request relates to issue [7195](https://github.com/jhipster/generator-jhipster/issues/7195).

For range filtering methods, I used **buildReferringEntitySpecification**, do you prefer **buildRangeReferringEntitySpecification** ? So it does not add this feature by default so user have to manually edit generated filter service classes to use the range filtering?

I used **root.get(reference)** for SingularAttribute but **root.join(reference)** for SetAttribute

@jdubois @gzsombor @DanielFran

### DIFF
--- a/jhipster-server/src/main/java/io/github/jhipster/service/QueryService.java
+++ b/jhipster-server/src/main/java/io/github/jhipster/service/QueryService.java
@@ -19,15 +19,17 @@
 
 package io.github.jhipster.service;
 
-import java.util.Collection;
-import javax.persistence.criteria.CriteriaBuilder.In;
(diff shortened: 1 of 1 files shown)


gzsombor:
Sorry for being late. This looks ok, however a bit complex and scary. But I suppose this is inevitable when someone tries to use JPA criteria API in a generic way :wink: 
 If you could come up with some elegant way, so this QueryService could be simplified a bit, please do not hold it back :)

PR #55, opened 2018-04-09
Remove spring-platform

erikkemperman:
Don't merge this yet, couchbase sample failing. See https://github.com/jhipster/generator-jhipster/issues/7422#issuecomment-379789335.

### DIFF
--- a/jhipster-server/pom.xml
+++ b/jhipster-server/pom.xml
@@ -214,17 +214,6 @@
     <build>
         <plugins>
             <plugin>
-                <groupId>org.sonatype.plugins</groupId>
-                <artifactId>nexus-staging-maven-plugin</artifactId>
-                <version>${nexus-staging-maven-plugin.version}</version>
-                <extensions>true</extensions>
-                <configuration>
-                    <serverId>ossrh</serverId>
-                    <nexusUrl>https://oss.sonatype.org/</nexusUrl>
-                    <autoReleaseAfterClose>true</autoReleaseAfterClose>
-                </configuration>
-            </plugin>
-            <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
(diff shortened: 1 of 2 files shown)


jhipster-server/pom.xml
```diff
@@ -213,17 +213,6 @@
 
     <build>
         <plugins>
-            <plugin>
-                <groupId>org.sonatype.plugins</groupId>
```
jdubois: are you sure this is not needed anymore? I think I need this to publish snapshot releases

erikkemperman: The same config is in the parent:
https://github.com/jhipster/jhipster/blob/master/jhipster-dependencies/pom.xml#L839
And we just need to enable the plugin:
https://github.com/jhipster/jhipster/pull/55/files#diff-580af11172ef8dd3e17b50bdf019f094L250

jdubois:
@erikkemperman the issue you have with Couchbase is a NoSuchMethodError on the the Couchbase client lib - so that's just a versioning issue

erikkemperman:
Yeah, I saw that, but as far as I can tell there are only patch-level increments with this change (I diffed the output of `mvn dependency:tree` before and after) and I would expect a disappearing method to be a major change!

erikkemperman:
The problematic change is here:
https://github.com/couchbase/couchbase-java-client/commit/762c2dc4460720fc5d27719630694e8badc9cd10#diff-5c321e9c5d4bec4a286000165da7bbd8L293

I have a lot of other things to do today, but will try to find some time to investigate further!

erikkemperman:
@jdubois Fixed!

jdubois:
Awesome, let's do this!

jdubois:
Let's test a little bit, and then I'll deploy a new snapshot - all JHipster v5 users will get it tomorrow, as it's a snapshot :-)

PR #57, opened 2018-04-13
Use 'dependencies' property

erikkemperman:
@jdubois
Fix the Travis build. No need to re-release the serverside stuff, the deployed artifacts are all right. (Evidence: generator builds are passing).

Things are a bit messy with these Travis scripts, complicated dependencies . When things settle down a bit I'll have a look if that can't be improved somewhat.


jdubois:
Thanks so much!!

PR #58, opened 2018-04-19
Add BOM for AWS to ensure aligned dependencies

ggotti:
I was having issues with the Amazon SSM version(**1.11.312**) pulling in an older version of the core library(**1.11.251**), which was causing runtime issues in the ECS container. I wasn't able to determine how maven was resolving to the older library.

Based on the instruction in [this AWS blog post](https://aws.amazon.com/blogs/developer/managing-dependencies-with-aws-sdk-for-java-bill-of-materials-module-bom/), I have added a BOM to ensure there is no mismatch.

```
[INFO] +- com.amazonaws:aws-java-sdk-ssm:jar:1.11.312:compile
[INFO] |  +- com.amazonaws:aws-java-sdk-core:jar:1.11.251:compile (version managed from 1.11.312)
[INFO] |  |  +- org.apache.httpcomponents:httpclient:jar:4.5.5:compile (version managed from 4.5.2)
[INFO] |  |  |  +- org.apache.httpcomponents:httpcore:jar:4.4.9:compile
[INFO] |  |  |  \- commons-codec:commons-codec:jar:1.11:compile (version managed from 1.10)
[INFO] |  |  +- software.amazon.ion:ion-java:jar:1.0.2:compile
[INFO] |  |  +- (com.fasterxml.jackson.core:jackson-databind:jar:2.9.3:compile - version managed from 2.6.7.1; omitted for duplicate)
[INFO] |  |  +- com.fasterxml.jackson.dataformat:jackson-dataformat-cbor:jar:2.9.3:compile (version managed from 2.6.7)
[INFO] |  |  |  \- (com.fasterxml.jackson.core:jackson-core:jar:2.9.3:compile - omitted for duplicate)
[INFO] |  |  \- (joda-time:joda-time:jar:2.9.9:compile - version managed from 2.8.1; omitted for duplicate)
[INFO] |  \- com.amazonaws:jmespath-java:jar:1.11.251:compile (version managed from 1.11.312)
[INFO] |     \- (com.fasterxml.jackson.core:jackson-databind:jar:2.9.3:compile - version managed from 2.6.7.1; omitted for duplicate)
```
I have also updated the version.

### DIFF
--- a/jhipster-dependencies/pom.xml
+++ b/jhipster-dependencies/pom.xml
@@ -39,7 +39,7 @@
 
     <properties>
         <!-- Dependency versions -->
-        <aws-java-sdk-ssm.version>1.11.312</aws-java-sdk-ssm.version>
+        <aws-java-sdk.version>1.11.313</aws-java-sdk.version>
         <bucket4j.version>3.1.1</bucket4j.version>
         <cassandra-unit-spring.version>3.3.0.2</cassandra-unit-spring.version>
         <couchmove.version>1.0.1</couchmove.version>
@@ -178,11 +178,6 @@
                 <groupId>io.jsonwebtoken</groupId>
                 <artifactId>jjwt</artifactId>
                 <version>${jjwt.version}</version>
-            </dependency>
-            <dependency>
-                <groupId>com.amazonaws</groupId>
-                <artifactId>aws-java-sdk-ssm</artifactId>
-                <version>${aws-java-sdk-ssm.version}</version>
             </dependency>
             <dependency>
                 <groupId>com.github.sabomichal</groupId>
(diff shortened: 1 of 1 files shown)


jdubois:
- The build failed as the BOM doesn't seem available on Maven Central
- Anyway I'm a bit worried that everybody will need to download that BOM, even when they don't use AWS. Couldn't you just fix the 2 required library: same mechanism as before, but both for the SSM and the core lib?

ggotti:
Oops, my mistake.

@jdubois I'm not familiar with the BOM system, but I thought it was just for centrally managing versions, not actually downloading the dependencies defined within them. My issue with fixing the core version only is that if the Amazon SSM Library changes in the future, then we'll have a similar issue where there will be a mismatch between versions.

jdubois:
No you were correct it's just for defining the libs, but the bom itself wasn't present

PR #59, opened 2018-04-22
Specification need to check if the specified field is refered entity or just an atribute

duderoot:
Filtering over metamodel should create an outer join when query is
entity?referenceEntityId.specified=boolean

The specification which is build based on an attribute, needs to check if this attribute is a referencing entity or is normal attribute

This is new PR for the new server side lib jhipster-framework which is gone be part o the new version 5.0.0.  PR [#37](https://github.com/jhipster/jhipster/pull/37) is based on top of server side lib jhipster version 1.3.0. I do not know what is intent regarding acceptance of a patch for JHipster V4.

Fixes jhipster/generator-jhipster#6735

### DIFF
--- a/jhipster-framework/src/main/java/io/github/jhipster/service/QueryService.java
+++ b/jhipster-framework/src/main/java/io/github/jhipster/service/QueryService.java
@@ -27,6 +27,7 @@ import org.springframework.transaction.annotation.Transactional;
 
 import javax.persistence.criteria.CriteriaBuilder;
 import javax.persistence.criteria.CriteriaBuilder.In;
+import javax.persistence.criteria.JoinType;
 import javax.persistence.metamodel.SetAttribute;
 import javax.persistence.metamodel.SingularAttribute;
 import java.util.Collection;
@@ -296,13 +297,27 @@ public abstract class QueryService<ENTITY> {
 
     protected <X> Specification<ENTITY> byFieldSpecified(SingularAttribute<? super ENTITY, X> field, final boolean
         specified) {
-        return specified ? (root, query, builder) -> builder.isNotNull(root.get(field)) : (root, query, builder) ->
-            builder.isNull(root.get(field));
+        if(isFieldAnReferringEntity(field)) {
+            return byReferringEntitiesFieldSpecified(field, specified);
+        }else {
+            return specified ? (root, query, builder) -> builder.isNotNull(root.get(field)) : (root, query, builder) ->
+                builder.isNull(root.get(field));
+        }
+    }
+
+    private boolean isFieldAnReferringEntity(SingularAttribute<? super ENTITY, ?> field) {
+        return field.getType().getJavaType().isInstance(field.getJavaMember().getDeclaringClass()) ? false : true;
     }
 
     protected <X> Specification<ENTITY> byFieldSpecified(SetAttribute<ENTITY, X> field, final boolean specified) {
         return specified ? (root, query, builder) -> builder.isNotEmpty(root.get(field)) : (root, query, builder) ->
             builder.isEmpty(root.get(field));
+    }
+
+    protected <X> Specification<ENTITY> byReferringEntitiesFieldSpecified(SingularAttribute<? super ENTITY, X> field, final boolean
+        specified) {
+        return specified ? (root, query, builder) -> builder.isNotNull(root.join(field, JoinType.LEFT)) : (root, query, builder) ->
+            builder.isNull(root.join(field, JoinType.LEFT));
     }
 
     protected <X> Specification<ENTITY> valueIn(SingularAttribute<? super ENTITY, X> field, final Collection<X>
--- a/jhipster-framework/src/main/java/io/github/jhipster/service/QueryService.java
+++ b/jhipster-framework/src/main/java/io/github/jhipster/service/QueryService.java
@@ -298,7 +298,7 @@ public abstract class QueryService<ENTITY> {
     protected <X> Specification<ENTITY> byFieldSpecified(SingularAttribute<? super ENTITY, X> field, final boolean
         specified) {
         if(isFieldAnReferringEntity(field)) {
-            return byReferringEntitiesFieldSpecified(field, specified);
+            return byReferringEntitiyFieldSpecified(field, specified);
         }else {
             return specified ? (root, query, builder) -> builder.isNotNull(root.get(field)) : (root, query, builder) ->
                 builder.isNull(root.get(field));
@@ -314,7 +314,7 @@ public abstract class QueryService<ENTITY> {
             builder.isEmpty(root.get(field));
     }
 
-    protected <X> Specification<ENTITY> byReferringEntitiesFieldSpecified(SingularAttribute<? super ENTITY, X> field, final boolean
+    protected <X> Specification<ENTITY> byReferringEntitiyFieldSpecified(SingularAttribute<? super ENTITY, X> field, final boolean
         specified) {
         return specified ? (root, query, builder) -> builder.isNotNull(root.join(field, JoinType.LEFT)) : (root, query, builder) ->
             builder.isNull(root.join(field, JoinType.LEFT));
--- a/jhipster-framework/src/main/java/io/github/jhipster/service/QueryService.java
+++ b/jhipster-framework/src/main/java/io/github/jhipster/service/QueryService.java
@@ -297,16 +297,12 @@ public abstract class QueryService<ENTITY> {
 
     protected <X> Specification<ENTITY> byFieldSpecified(SingularAttribute<? super ENTITY, X> field, final boolean
         specified) {
-        if(isFieldAnReferringEntity(field)) {
+        if(field.isAssociation()) {
             return byReferringEntitiyFieldSpecified(field, specified);
         }else {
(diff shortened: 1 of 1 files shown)


gzsombor:
As the library is used by code generation, reflection wouldn't be needed. Just provide a method for _plain old types_ and a separate method for _entities_

Blackdread:
Is left join really necessary in both cases (for isNotNull and isNull)? No risk of getting too many results in certain case not desired (instead of inner join)?

@gzsombor You mean that method "isFieldAnReferringEntity" is not required?

Suggestion:
For consistency with all methods using singular "buildReferringEntitySpecification" for entity word, maybe can change "byReferringEntitiesFieldSpecified" to "byReferringEntityFieldSpecified" or "byFieldSpecifiedReferringEntity" (second choice better as in IDE it will still show both methods when already wrote 3 first characters).

jhipster-framework/src/main/java/io/github/jhipster/service/QueryService.java
```diff
@@ -297,24 +297,20 @@
...
-        return field.getType().getJavaType().isInstance(field.getJavaMember().getDeclaringClass()) ? false : true;
-    }
-
     protected <X> Specification<ENTITY> byFieldSpecified(SetAttribute<ENTITY, X> field, final boolean specified) {
         return specified ? (root, query, builder) -> builder.isNotEmpty(root.get(field)) : (root, query, builder) ->
             builder.isEmpty(root.get(field));
     }
 
-    protected <X> Specification<ENTITY> byReferringEntitiesFieldSpecified(SingularAttribute<? super ENTITY, X> field, final boolean
+    protected <X> Specification<ENTITY> byReferringEntitiyFieldSpecified(SingularAttribute<? super ENTITY, X> field, final boolean
```
Blackdread: typo Entitiy - Entity

Blackdread:
I think gzombor meant that your if/else can be removed in the method "byFieldSpecified" as QueryService already have "buildSpecification" and "buildReferringEntitySpecification" (and "buildRangeSpecification" and "buildStringSpecification" etc) so distinction between owner side and non-owner side of your use case of OneToOne is solved via that. You might be able to replace the call of "byFieldSpecified" in methods "buildReferringEntitySpecification" (I have not check but that's what I interpret from that.

duderoot:
@Blackdread thank you for the hints. I make some tests locally and it's looks ok. 
What is the strategy for making test for this code. From what I see it's look like the generator is creating tests classes for this code when using the `jpaMetamodelFiltering` option. It looks to me that there is no test for reference entity specified. Maybe when this is closed we can add some test to the generator.

Blackdread:
I think you missed the other method "buildReferringEntitySpecification(Filter<X> filter, SingularAttribute<>...", you only modified "buildReferringEntitySpecification(final RangeFilter<X> filter, SingularAttribute<>...".


I also do not understand why you used "JoinType.LEFT" instead of just "root.join(field)", the left join might return too many results.

This is for another discussion but we could also create:
```
protected <OTHER, X> Specification<ENTITY> byReferringEntityFieldSpecified(SingularAttribute<? super ENTITY, OTHER> reference, SingularAttribute<OTHER, X> otherField, final boolean specified) {
        
return specified ? 
  (root, query, builder) -> builder.isNotNull(root.join(reference).get(otherField)) : 
  (root, query, builder) -> builder.isNull(root.join(reference).get(otherField));
}
```
where I replaced "idField" by "otherField" because we could filter not only on id of the relation (this is what is generated by default -> only id). It would require to review other methods, change "idField" and make criterias even bigger to be able to filter on more fields (maybe too much)...

gzsombor:
Yes, that's what I wanted to suggest. One more question - which came to my mind reading the code again - what happened with the foreign key? A foreign key should be there to validate that the id which is stored points to a proper, and existing record in the other table.

gzsombor:
I think, this is already solved.

duderoot:
@gzsombor I am affaraid that this did not solved the problem that we try to resolve in [generator-jhipster/#6735](https://github.com/jhipster/generator-jhipster/pull/6735). @jdubois had try to solve this in [generator-jhipster/#8411](https://github.com/jhipster/generator-jhipster/pull/8411) and @kaidohallik als confirmed that [generator-jhipster/#6735](https://github.com/jhipster/generator-jhipster/pull/6735) is still not fixed by the latest master  version in jhipster-framework. 
As a solution , to stay in the context of [generator-jhipster/#8411](https://github.com/jhipster/generator-jhipster/pull/8411)), we need a posibility to build a query which can filter all the Sponsor which do not have a Sponsoragreemant like 
`http://localhost:8080/api/sponsors?filter=sponsoragreement-is-null`
 
The only way I manged to create a filter over the QueryService that produce the expected result is by adding a new method like:

```java
protected <X> Specification<ENTITY> buildReferringEntityFieldSpecified(SingularAttribute<? super ENTITY, X> field, final boolean
            specified) {
            return specified ? (root, query, builder) -> builder.isNotNull(root.join(field, JoinType.LEFT)) : (root, query, builder) ->
                builder.isNull(root.join(field, JoinType.LEFT));
        }
```


and then creating a Specification in SponsorQueryService to call the QueryService like this:

```java
specification.and(buildReferringEntityFieldSpecified(Sponsor_.sponsorAgreement, false));
```
Can you please give me a hint on how I can build a queery like buildReferringEntityFieldSpecified or if its not possible reopen this issue so I can submit a new PR. thank you

duderoot:
@Blackdread thank you for your answer. I have no problem in changing the generated code. Then I can create a PR for the generator. My problem is that I don't quite get it how to use the QueryService to  generate a Specification like the one that I need e.g. `buildReferringEntityFieldSpecified`. Can you please be a little more specific and give me an example on how to create a Specification in the SponsorService using the new QueryService? Thank you.

gzsombor:
I think, you need to write:

```
Filter<Long> sponsorFilter = ...
spec = spec.and(buildReferringEntitySpecification(sponsorFilter, 
    root -> root.join(Sponsor_.sponsorAgreement), 
    entity -> entity.get(SponsorAgreement_.id) ));
```

kaidohallik:
In https://github.com/jhipster/jhipster/blob/master/jhipster-framework/src/main/java/io/github/jhipster/service/QueryService.java there are currently:
* 2 methods of buildReferringEntitySpecification for one-to-one and many-to-one (one is old deprecated method)
* 4 methods of buildReferringEntitySpecification for one-to-many and many-to-many

For one-to-one and many-to-one there is currently missing method which respects Join as parameter.

gzsombor:
Yes, it's a bit confusing, and more function can be deprecated to reduce the clutter.
I think, you can use `buildSpecification(Filter<X> filter, Function<Root<ENTITY>, Expression<X>> metaclassFunction)`, just like as in https://github.com/jhipster/jhipster/blob/master/jhipster-framework/src/main/java/io/github/jhipster/service/QueryService.java#L189 - where the  old style buildReferringEntitySpecification is implemented with the new lambda-style function.

Blackdread:
To have 'cross join' is bad, can you show the 'metaclassFunction' that you passed?
Which part is the owner side (has the foreign key in his table)?

kaidohallik:
I made some tests with these rows (identical function bodies):
* https://github.com/jhipster/jhipster/blob/master/jhipster-framework/src/main/java/io/github/jhipster/service/QueryService.java#L189
* https://github.com/jhipster/jhipster/blob/master/jhipster-framework/src/main/java/io/github/jhipster/service/QueryService.java#L217

Input filter: `specified=false`

Current implementation returns correct resultset in owner side and always empty resultset in no owner side:
```
        return buildSpecification(filter, root -> root.get(reference).get(valueField));
```

Changing this to following returns always empty resultset in both owner and no owner side:
```
        return buildSpecification(filter, root -> root.join(reference).get(valueField));
```

Changing this to following returns correct resultset both in owner and no owner side:
```
        return buildSpecification(filter, root -> root.join(reference, JoinType.LEFT).get(valueField));
```

duderoot:
@Blackdread  to answer your question:

> To have 'cross join' is bad, can you show the 'metaclassFunction' that you passed?
> Which part is the owner side (has the foreign key in his table)?

SponsorAgreement is the owner.
I guess I found what the problem was and also hope to have a solution for this. The [PR#8449](https://github.com/jhipster/generator-jhipster/pull/8449) is changing the call from the SponsorQueryService so that now when the API is sending a call like
`http://localhost:8080/api/sponsors?sponsorAgreementId.specified=true`
to the BE.
From the BE the QueryService
```
byFieldSpecified(Function<Root<ENTITY>, Expression<X>> metaclassFunction,
            final boolean specified)
```
is getting caled  from SponsrQueryService like this:
````
buildSpecification(criteria.getSponsorAgreementId(),
                        root -> root.join(Sponsor_.sponsorAgreement, JoinType.LEFT).get(SponsorAgreement_.id))
````

It should not break anything else. For me it feels a little strange that there is no test coverage in QueryService and I am facing some problems to understand  all the code in there.
@kaidohallik thank you for confirming my assumption. It makes my fell not to be to far from a solution or understanding of a problem 😄

gzsombor:
Yes, unfortunately no real testing for the QueryService, however, in the generated code, there are tests for each attribute - from the Controller/REST layer - unfortunately no relation testing happens as of yet. 
 It would be nice :)

PR #67, opened 2018-05-17
Prevent thread lock with spring-cloud-context

juliensadaoui:
Fix #7478

### DIFF
--- a/jhipster-framework/src/main/java/io/github/jhipster/config/liquibase/AsyncSpringLiquibase.java
+++ b/jhipster-framework/src/main/java/io/github/jhipster/config/liquibase/AsyncSpringLiquibase.java
@@ -33,6 +33,9 @@ import org.springframework.util.StopWatch;
 import liquibase.exception.LiquibaseException;
 import liquibase.integration.spring.SpringLiquibase;
 
+import java.sql.Connection;
+import java.sql.SQLException;
+
 /**
  * Specific liquibase.integration.spring.SpringLiquibase that will update the database asynchronously. <p> By default,
  * this asynchronous version only works when using the "dev" profile.<p> The standard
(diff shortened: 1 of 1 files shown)


jhipster-framework/src/main/java/io/github/jhipster/config/liquibase/AsyncSpringLiquibase.java
```diff
@@ -71,14 +74,20 @@ public AsyncSpringLiquibase(@Qualifier("taskExecutor") TaskExecutor taskExecutor
...
-                    try {
-                        logger.warn(STARTING_ASYNC_MESSAGE);
-                        initDb();
-                    } catch (LiquibaseException e) {
-                        logger.error(EXCEPTION_MESSAGE, e.getMessage(), e);
-                    }
-                });
+                // Prevent Thread Lock with spring-cloud-context GenericScope
+                // https://github.com/spring-cloud/spring-cloud-commons/commit/aaa7288bae3bb4d6fdbef1041691223238d77b7b#diff-afa0715eafc2b0154475fe672dab70e4R328
+                try (Connection connection = getDataSource().getConnection()) {
```
jdubois: Don't you need to release/close the connection at the end? I'm not sure how it works with the new `try()` blocks

juliensadaoui: The class java.sql.Connection implements the interface AutoCloseable since JDK 7 and can be used as resources in a try-with-resources statement.

jdubois: Awesome, thanks!

PR #68, opened 2018-05-22
Create referring entity specification with inheritance.

Earthmor:
Fixed situation when reference and valueField have a common ancestor but valueField is located in the superclass.
Gist with problems example - https://gist.github.com/Earthmor/21eb813c14bdd5325c9b575615070a5e


gzsombor:
I've verified it, and this is indeed the correct solution for   jhipster/generator-jhipster#8040.

I'm going to add a 'compile test' for this too

PR #70, opened 2018-05-29
Bump dependencies

erikkemperman:
A few minor bumps, and one major: We were several years out of date, having missed that the `groupId` of cucumber artifacts had changed :-)

@jdubois
I would also like to propose keeping the versions of `jhipster-dependencies` and `-framework` in both this repo as well as the generator on a SNAPSHOT -- and only change it to "real" versions just before a release. That way we don't use up quite so many minor versions which only serve to communicate changes to build/develop iterations in between "proper" releases.

Does that make sense, or have been misunderstanding your intentions here?

Cf., this depends on [this generator PR](https://github.com/jhipster/generator-jhipster/pull/7702).

I'll cancel Travis: it will fail because of mutually dependent changes, see here for passing tests where both generator and server-lib PRs are applied:

https://travis-ci.org/erikkemperman/generator-jhipster/builds/385069631
https://travis-ci.org/erikkemperman/jhipster/builds/385070081


jdubois:
OK this all makes sense to me, let's try this

pascalgrimaud:
I restarted all builds this morning.
Everything is green!

@erikkemperman : FYI, Spring Cloud Finchley RC2 has been released yesterday: https://spring.io/blog/2018/05/29/spring-cloud-finchley-rc2-has-been-released

> Finchley.RC2 is based on Spring Boot 2.0.2.

I think we should upgrade. What do you think?

erikkemperman:
@pascalgrimaud Damn! I checked a couple of times the last few days -- including just before I pushed my branch, at that point it was something like 9 days overdue, but I didn't want to wait for it. Yeah, I'll make a new PR, we definitely want RC2 (RELEASE is scheduled for June 13, but it might get delayed as well, probably that shouldn't block JH5).

About those failing Travis builds, I checked a few logs and it seems the Yarn registry is unreliable. I noticed that before. Do we still want to default to Yarn? NPM has caught up, pretty much. In any case, maybe we should prefer the NPM registry even with Yarn client?

https://yarnpkg.com/en/docs/cli/config/

jdubois:
Yes let's upgrade

erikkemperman:
@jdubois Done, https://github.com/jhipster/jhipster/pull/72

PR #71, opened 2018-05-29
Use a simple utility class for time-limited token cache

erikkemperman:
I'd like to get rid of the huge Guava dependency, which we only use for two "small" purposes that are easily replaced.

This is a simple custom cache for tokens, see https://github.com/jhipster/generator-jhipster/issues/6648#issuecomment-345823580 for original discussion. I've incorporated the suggestions by caffeine dev Ben Manes, and moved the cache to server lib as requested.

My impression is that for this particular scenario a custom implementation might actually be better than using the Spring Cache abstraction, I looked briefly but couldn't find a setup which uses the special circumstances to make the lookups and maintainance as fast. But we can always go that way later of course, my goal here is primarily to ditch Guava.

This has a counterpart PR on the generator, [here]().

I will cancel failing Travis due to this mutual dependence, see here for passing tests with both PRs applied:
https://travis-ci.org/erikkemperman/generator-jhipster/builds/385181814
https://travis-ci.org/erikkemperman/jhipster/builds/385181785


erikkemperman:
@jdubois Sorry, I guess maybe I should have said that I was planning on making another PR -- now it crossed with your release. This is why snapshots are nice :-)

jhipster-framework/src/main/java/io/github/jhipster/security/PersistentTokenCache.java
```diff
@@ -0,0 +1,128 @@
...
+ * Simple time-limited cache for login tokens, necessary to avoid concurrent
+ * requests invalidating one another. It uses a {@link java.util.LinkedHashMap}
+ * to keep the tokens in order of expiration. During access any entries which
+ * have expired are automatically purged.
+ */
+public class PersistentTokenCache<T> {
+
+    private final long expireMillis;
+
+    private final LinkedHashMap<String, Value> map;
```
MathieuAA: Can't it be `private final Map<String, Value> map;`?

erikkemperman: Sure, with an extra `import` (or asterisk). Would that be preferable? The fact that it's `Linked` is relevant to this implementation, so I thought I would make that explicit also in the field type.

MathieuAA: It's your call Eric, I just prefer not to have a concrete implementation when declaring fields.

jdubois: I've always been taught you should use the interface here, so let's do a Map, at least for consistency with the rest of the code. Nobody will see this anyway as it's private, so it's no big deal.

erikkemperman:
All right, done!

erikkemperman:
Hold on, something is going wrong -- let me check...

erikkemperman:
Ugh, had to keep the managed Guava dependency after all -- otherwise springfox/swagger uses an ancient version and its `Function` class doesn't inherit from the new standard Java `Function`. So I propose to keep the managed Guava in here, so we can cast.

Note that this doesn't invalidate this PR, imho, I would still like to have the option to not use Guava in my generated apps (I don't use swagger in production builds).

By the way, SpringFox 2.9.0 is released but not available on Maven... https://github.com/springfox/springfox/releases

erikkemperman:
Just to avoid confusion, I don’t think this PR has to wait for the next Springfox. See [rationale](https://github.com/jhipster/jhipster/pull/72#issuecomment-393434285).

erikkemperman:
@jdubois Thanks Julien! When this is released, the accompanying PR on the generator can be considered as well:
https://github.com/jhipster/generator-jhipster/pull/7704

PR #72, opened 2018-05-30
Bump spring-cloud to Finchley.RC2

erikkemperman:
Cf., https://github.com/jhipster/jhipster/pull/70#issuecomment-393045121

### DIFF
--- a/jhipster-dependencies/pom.xml
+++ b/jhipster-dependencies/pom.xml
@@ -64,7 +64,7 @@
         <problem-spring-web.version>0.23.0</problem-spring-web.version>
         <prometheus-simpleclient.version>0.4.0</prometheus-simpleclient.version>
         <reflections.version>0.9.11</reflections.version>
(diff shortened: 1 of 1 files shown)


jdubois:
OK I'm releasing a patch version for this

erikkemperman:
Did you take a look at my other PR [here](https://github.com/jhipster/jhipster/pull/71)? If you don't like it or it needs improvement that's fine of course -- but otherwise perhaps it could be included in the release?

erikkemperman:
Ah, nevermind, you did the release already. Sorry, I could have checked before I posted :)

jdubois:
Yes of course, and it would be really good to get rid of Guava for "normal" users. But I saw we needed to wait for the next SpringFox release, so I'm waiting.

jdubois:
yes @erikkemperman I totally agree, no problem :-)
and that's great :-)

PR #73, opened 2018-06-02
fix couchbase-client version to 2.5.7

JulioJu:
See https://github.com/jhipster/generator-jhipster/issues/7654, escpecially https://github.com/jhipster/generator-jhipster/issues/7654#issuecomment-394086971

@erikkemperman, @tchlyah therefore  db4b2ba0981604b8c490dd627b9245aa40f3bb1d should be reverted ?


erikkemperman:
Yeah, ok, I had missed the comment that 2.5.8 would not work. Sorry about that. Glad I do the upgrades in little individual commits :-)

@jdubois Could you revert this commit? I’m traveling..

PR #83, opened 2018-07-30
Update spring-boot to 2.0.4, spring-cloud to Finchley.SR1 and other dependencies

DanielFran:


DanielFran:
I rollback the update of couchbase-client because it might need the update to Couchbase Server 5.5 and I do not have the skills to test it.

DanielFran:
@tchlyah Maybe you can take a look after for couchbase?

PierreBesson:
For me this is good. But it will require a release of the lib. So let's wait for @jdubois to come back from vacation. No hurry.

PR #84, opened 2018-07-31
Add util query and criteria

Blackdread:
See issue [#7574](https://github.com/jhipster/generator-jhipster/issues/7574)

I have implemented only with joins so less code (anyway only for fk column we could skip a join but that does not impact on performance).

Not to be merged yet because QueryUtil can still define more functionality like SetAttribute.
CriteriaUtil is not complete, it just have basic example of some useful methods but a lot more can be added (not all type of Filter put, only equals and IN for now, etc)

CriteriaUtil is an opiniated implementation (some are used in my projects) but can always be overridden/hidden if required

NOTE: duplicate from https://github.com/jhipster/jhipster/pull/64 but I messed up again the history of commits...

### DIFF
--- a/jhipster-framework/src/main/java/io/github/jhipster/service/util/CriteriaUtil.java
+++ b/jhipster-framework/src/main/java/io/github/jhipster/service/util/CriteriaUtil.java
@@ -1,0 +1,1797 @@
+package io.github.jhipster.service.util;
+
+import io.github.jhipster.service.filter.*;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.Nullable;
+import javax.validation.constraints.NotNull;
+import java.math.BigDecimal;
+import java.time.Instant;
+import java.time.LocalDate;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Allow to build or erase parts of criteria received in REST API, often we want to make sure that for getAll or getOne we filter on specific data that user can only get and not others.
+ * <p>The logic follow: equals &gt; in &gt; contains &gt; specified = gt = gte = lt = lte. It is similar to order of {@linkplain io.github.jhipster.service.QueryService}</p>
+ * <p>If further checks are required like when building equals criteria that "IN" must be empty or only contain the equal value passed, you can extends this class.</p>
+ * <p>Some methods may be "useless" but it is provided as is, user can decide to use or not.</p>
+ * <p>Created on 2018/2/6.</p>
+ *
+ * @author Yoann CAPLAIN
+ */
+@SuppressWarnings({"unused"})
+public class CriteriaUtil {
+
+    private static final Logger log = LoggerFactory.getLogger(CriteriaUtil.class);
+
+    protected CriteriaUtil() {
+    }
+
+    protected static <T extends Filter<?>> T createCriteria(final Class<T> criteriaClass) {
+        try {
+            return criteriaClass.newInstance();
+        } catch (InstantiationException e) {
+            log.error("Make sure that filters passed can have a default empty constructor", e);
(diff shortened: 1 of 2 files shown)


gzsombor:
This is huge and complicated. What I don't understand, why reflection is needed to construct criteria objects? I don't get, why someone would want to call buildCriteria(StringFilter.class. null, ... ), instead of buildCriteria(new StringFilter(),... )
The other confusing thing are the methods with 7-8 generic parameters. For example, Instead of

    public static <U, ENTITY1, ENTITY2, ENTITY3, ENTITY4> Specification<ENTITY1> buildEqualsJoin(final U value, final SingularAttribute<? super ENTITY1, ENTITY2> e1, final SingularAttribute<? super ENTITY2, ENTITY3> e2, final SingularAttribute<? super ENTITY3, ENTITY4> e3, final SingularAttribute<? super ENTITY4, U> e4) {
        return (root, query, builder) -> builder.equal(root.join(e1).join(e2).join(e3).get(e4), value);
    }

You could have a method

    public static <U, ENTITY> Specification<ENTITY1> buildEqualsJoin(final U value, final Function<Root<ENTITY>, U> attributePicker) {
        return (root, query, builder) -> builder.equal(attributePicker.apply(root), value);
    }


And you can call it from your hand-made code like this:

    buildEqualsJoin(currentValue, (root) -> root.join(MyObject_.field).join(OtherObject_.other).get(ThirdObject_.field));

This would make a bit simpler, I hope

 having `final SingularAttribute<? super ENTITY1, ENTITY2> e1, final SingularAttribute<? super ENTITY2, ENTITY3> e2, final SingularAttribute<? super ENTITY3, ENTITY4> e3, final SingularAttribute<? super ENTITY4,U> e4`, you could have a method, which takes a function parameter,

Blackdread:
I agree that no one would want to use reflection and my methods buildXXXXX(final Class<T> class, ... ), it is mostly for internal use (I never use outside), I made it public for examples in my pull request.
All build methods are overloaded to provide easy creation by passing directly the filter instance.


Should not your code be : 
```
public static <ENTITY, U> Specification<ENTITY> buildEqualsJoin(final U value, final Function<Root<ENTITY>, Path<U>> attributePicker) {
        return (root, query, builder) -> builder.equal(attributePicker.apply(root), value);
    }
```
I like your new code with the use of Function<...>, and it solves the issue that currently I only support SingularAttribute join (not PluralAttribute).
The only problem I see is, it makes more code to write with your method :P ^^ but it is open to a lot more possibilities with SingularAttribute / PluralAttribute / etc.


So far I used my code to do things like (in QueryService implementations):
```
specification = specification.and(QueryUtil.buildReferringEntitySpecification(criteria.getCompanyId(), MyObject_.field, OtherObject_.company, Company_.id));
```


Other than that, do you think we could add util class like QueryUtil and CriteriaUtil into jHipster? (maybe not exactly my implementation of CriteriaUtil as we can solve the 7-8 generics with only the Function<>)

gzsombor:
The problem with generics, and all this meta-class-machinery is that no-one really understands it :wink: 
I think, the problem, which you try to solve is a real one, worth providing a solution. Fortunately I haven't encountered a case, where 7 levels of foreign keys would need to be joined, but maybe I was just lucky. 
 The other way to make this filtering more generic is to add a way to put criteria classes inside each other. So for example an OrderCriteria could contain a ItemCriteria, which contains a PartCriteria. So you could use the REST interface like 'order.line.part.equals=5323'. I bet, it would be possible to generate everything, but the complexity and maintenance would be very difficult. (And there is 'GraphQL' thingy already out there, which is claimed to be a 'standard', to solve similar problems)

gzsombor:
I think, we should close this issue, as a different solution has been merged.

PR #85, opened 2018-07-31
Add equals and hashcode for filters.

Blackdread:
See [#8049](https://github.com/jhipster/generator-jhipster/issues/8049)

Put some test as well


gzsombor:
Could you put braces, and line brakes in the "if"-s. It makes it harder to read. Other than, it's a good idea !

Blackdread:
Tests pass :) Can merge if all good for you :)

PR #87, opened 2018-08-21
Add Spring Data Jest version

mraible:
I'm not sure I'm doing this right, but here's an attempt to add a version for Spring Data Jest that's added by https://github.com/jhipster/generator-jhipster/pull/8122.


mraible:
It'd be cool if we could exclude commons-logging from `io.searchbox:jest` as well. Can that be done with these projects?

jdubois:
@mraible the JJWT upgrade isn't working well as the artifactId changed, see https://search.maven.org/search?q=jjwt 
As this is a different matter, could you split your PR in two? So we handle the JJWT issue separately

mraible:
@jdubois I changed this PR to only add spring-data-jest. I'll open a new PR for the JJWT upgrade.

PR #89, opened 2018-08-22
Upgrade JJWT to 0.10.5

mraible:


jdubois:
This makes the build fails as this also depends on jhipster/generator-jhipster -> that's normal, let me do the release, I'll probably break the build for a couple of hours, but I'll fix everything myself

PR #98, opened 2018-10-04
pin dependency version to fix some cves

atomfrede:
closes https://github.com/jhipster/generator-jhipster/issues/8248


jdubois:
Oh OK @atomfrede so not really a CVE then! I hope this doesn't cause any other issue, I'm trying to do a release today, so it's in the next JHipster patch (also due today)

atomfrede:
And when again thinking of it it might be irrelevant for us: `In Apache Log4j 2.x before 2.8.2, when using the TCP socket server or UDP socket server to receive serialized log events from another application, a specially crafted binary payload can be sent that, when deserialized, can execute arbitrary code.`

https://nvd.nist.gov/vuln/detail/CVE-2017-5645

PR #102, opened 2018-10-06
configures maven pom to produces reports and releases

Tcharl:
Enables some nice capabilities to Jhipster repo maintainers:

- Ability to release a snapshot pom using these two commands
```
./mvnw release:prepare --batch-mode -Prelease -s /.../settings.xml
./mvnw release:perform --batch-mode -Prelease -s /.../settings.xml
```
This release will increment SNAPSHOT to version to SNAPSHOT+1, tag release version.

- Generates a site listing dependencies or plugins upgrade (mvn site, or on the gh-pages branch after a release), which could be useful for maintaining bom

- Generates License headers automatically (that's the reason of all files modification).

Only the parent pom has been edited, and a new src/site/site.xml file at the root

### DIFF
--- a/jhipster-framework/LICENSE.txt
+++ b/jhipster-framework/LICENSE.txt
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
+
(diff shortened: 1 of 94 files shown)


DanielFran:
@Tcharl Let's confirm if @jdubois agree with that?

PierreBesson:
+1 for me !

Tcharl:
I can't be a 100% sure for the release, as I haven't the key and don't want to release jhipster, it could need some tweak, but this should be light, I already use this configuration on many of my projects.

jdubois:
I added a review as there is clearly an error with one license file, and this should not be merged.

Otherwise, I'm really not a big fan, as we lose the correct license headers in all our files. I don't want automatic header licenses, I want the correct license in the source code from the start, so when someone clones the project he has the license everywhere. Otherwise it's too easy for people to say they cloned the project, so they don't have to respect the license.

I'm sorry to be annoying here, but licenses are the most important thing we have, and I have no problem to do a release at the moment, I just upgrade the version number and `mvn clean deploy`.

PierreBesson:
@jdubois, I don't understand. I thought the licence obviously apply to every file of the repo because of the LICENSE file at the root of the repo.

For me the license headers are only a convenience to know the licensing information when downloading the sources in the IDE. Also it is valuable to have a tool to add them for you (most contributors initially forget to put them when they add new classes).

PierreBesson:
Apart from this I agree that the header we had before was clearer than the proposed one.

Tcharl:
So, if I understand well, the blocker here is the license header, I can modify it to fit the original one...

jdubois:
@Tcharl yes, for the rest:

- it generates a site with some information, I guess some people can find this useful? Issue for me: I have no idea how to host this, not sure it can be done with the release notes (as it's already a huge website and we are already way above the GitHub quotas)
- it provides some automatic release commands: if you can document them, then we could indeed use them easily

Tcharl:
@jdubois ,

- For the site, you do not have to host it: github does it automatically for you when pushing to the gh-pages branch. Also, it is made for us (the maintainers) in order to easily check dependency updates and convergence.

- For the release command: of course I should document it.

PR #104, opened 2018-10-11
add feign pageable configuration

Tcharl:
see https://github.com/jhipster/generator-jhipster/pull/8480

### DIFF
--- a/jhipster-framework/pom.xml
+++ b/jhipster-framework/pom.xml
@@ -157,6 +157,11 @@
             <optional>true</optional>
         </dependency>
         <dependency>
+            <groupId>org.springframework.cloud</groupId>
+            <artifactId>spring-cloud-starter-openfeign</artifactId>
+            <optional>true</optional>
+        </dependency>
+        <dependency>
             <groupId>org.springframework.security</groupId>
             <artifactId>spring-security-data</artifactId>
             <optional>true</optional>
--- a/jhipster-framework/src/main/java/io/github/jhipster/config/feign/PageableQueryEncoder.java
+++ b/jhipster-framework/src/main/java/io/github/jhipster/config/feign/PageableQueryEncoder.java
@@ -1,0 +1,59 @@
+package io.github.jhipster.config.feign;
+
+import feign.RequestTemplate;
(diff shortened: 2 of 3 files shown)


DanielFran:
@Tcharl @PierreBesson  Why not use the library itself instead? https://github.com/ElderByte-/java-starter

PierreBesson:
@DanielFran Yes this project is very cool but it's a spring boot starter so it will force us to use their config for exemple these hardcoded properties: https://github.com/ElderByte-/java-starter/blob/master/starter-spring-cloud/src/main/java/com/elderbyte/spring/cloud/bootstrap/feign/DefaultEnvironmentConfigurer.java#L38.

What would be great would be if @IsNull 's project could split the utils like the pageable in a new library, then we could import only those classes. @IsNull, would you be interested to work together on this ? I don't like that this solution is imposing work on you to publish a library to maven central just for a few classes so the simplest solution for us is still to copy your code.

DanielFran:
@PierreBesson you are right, I did not check that it was mandatory to use the starter.
In fact what I understand from @IsNull was that he was working on separate the project in different libraries.... maybe I was confused, but his feedback was deleted with the last refactor from @Tcharl .

IsNull:
Well ideally this should be supported by spring out of the box, and this is why I opened a [PR on spring boot](https://github.com/spring-cloud/spring-cloud-netflix/pull/1604) directly. While the implementation worked and was tested, it was still not enough unit tests for them and I was not able to spend more time on it, thus decided to put this into our own starter lib for a quick solution.

Long story short, best would be to make a new PR directly on spring. Its a feature which I have seen requested a lot.

jdubois:
Yes I'd rather have the solution from @IsNull

Tcharl:
[Another try](https://github.com/spring-cloud/spring-cloud-openfeign/pull/103)

DanielFran:
@Tcharl Since we upgrade to Spring Cloud Greenwish SR1, will you review this PR to implement pageable?

Tcharl:
No Need to have server code anymore, closed in favor of the generator-jhipster one

PR #135, opened 2018-12-04
Force ElasticSearch version

avdev4j:
Spring boot POM dependencies set ElasticSearch 6.4.3 but we are using spring data jest and spring data elasticSearch that need ElasticSearch 6.2.2.
We can't use ElasticSearch 6.4.3 yet because the move some method from XContentBuilder and it's break at server startup.

Maybe we should remove it when spring data jest will handle 6.3+

#8683


DanielFran:
@avdev4j I open issue https://github.com/VanRoy/spring-data-jest/issues/102

PR #142, opened 2018-12-26
update sb-2.1.1

srinivasa-vasu:
- update infinispan sb version

infinispan 2.1.1 seems to have some issues with actuator integration while using non-native cache like JCache


DanielFran:
@srinivasa-vasu I updated now with new version 2.1.2 that include your fix!

PR #144, opened 2018-12-30
configures maven release plugin

Tcharl:
[Maven release without license plugin](https://github.com/jhipster/jhipster/pull/102 )

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -16,6 +16,7 @@
     <name>JHipster server-side parent POM</name>
     <url>https://github.com/jhipster/jhipster/</url>
 
+    <inceptionYear>2016</inceptionYear>
     <licenses>
         <license>
             <name>Apache License, version 2.0</name>
@@ -47,8 +48,10 @@
         <!-- Build properties -->
         <java.version>1.8</java.version>
         <maven.version>3.0.0</maven.version>
-        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
+        <encoding>UTF-8</encoding>
+        <project.build.sourceEncoding>${encoding}</project.build.sourceEncoding>
+        <project.reporting.outputEncoding>${encoding}</project.reporting.outputEncoding>
+
         <project.testresult.directory>${project.build.directory}/test-results</project.testresult.directory>
         <maven.build.timestamp.format>yyyyMMddHHmmss</maven.build.timestamp.format>
         <maven.compiler.source>${java.version}</maven.compiler.source>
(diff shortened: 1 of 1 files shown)


DanielFran:
@jdubois Maybe we can include this in spring-boot 2.1 branch and you can test the new process to release the BOM?

jdubois:
Is this only for doing new releases of jhipster/jhipster? I'm sorry, but what does it add to what I do today, which is just "mvn clean deploy"? This has worked very well for quite some time, so I'm reluctant to change without any reason.

DanielFran:
@jdubois From the previous PR it seems that the blocker was the licence files generated, and now @Tcharl removed this part...
So now, maybe this PR make no sence!?!

jdubois:
@DanielFran oh sorry, I need to read that PR again, and I need to find it back :-)

DanielFran:
@jdubois Can you please review this PR and the inicial one (https://github.com/jhipster/jhipster/pull/102) to confirm if it is still OK before requesting @Tcharl to fix the conflicts

jdubois:
- I'm not sure about the license part, I'm happy with the current manual license headers. Then, of course, that probably removes some copy/paste errors.
- For reports/website generation, maybe some users will like them. I don't know where we could host them, so those would mostly be for end-users. I'm not sure if people still use those?
- For the automatic releases, I'm adding @pascalgrimaud as reviewer, as he's doing them now (also still with me, hopefully!!).

jdubois:
So I've launched a Twitter poll for the "Maven site & reports" part: https://twitter.com/juliendubois/status/1167341636482994176

I know it's not always worth much, but the results are that most people don't really care, but we have a good percentage who find them "very useful", so I think we should have them.

I just have 2 questions:

- Would adding them change anything to people who don't want them? I'm guessing that's just a few extra lines of XML in the pom, and nothing else?? BTW we should see if we could modularize our pom.xml, I have no idea how this works, but I've just seen someone do this and that definitely cleans up the file (especially if you use a code generator).
- I don't think we would have an easy way to host those for our own projects, would that be an issue? People wouldn't have any difference from what they have now.

DanielFran:
Hello @jdubois, are we moving forward with this implementation?

jdubois:
@DanielFran I'm worried to add even more XML lines to our pom.xml for only a minority of people. It's already far too complex.

I've seen some people splitting their pom.xml into several files and then importing them, do you have any experience with that? Maybe we could refactor the whole pom.xml to make it much simpler and cleaner, and in that process having a specific file to generate the report wouldn't be so problematic

pascalgrimaud:
Agree with @jdubois 
Let's stay like this, we have other stuff with higher priority in the roadmap of v7

DanielFran:
@jdubois Splitting is doing like @Tcharl does in his module: https://github.com/Tcharl/generator-jhipster-enterprise-pom/tree/master/generators/app/templates/mvn

DanielFran:
@pascalgrimaud It is just to update status of the current PR in this project... implementing or closing them.

pascalgrimaud:
You're right we need to clean issue or PR, which are opened for too long.
So let's close this one.

jdubois:
Thanks @DanielFran - I've only seen this once on a project, so I'm not sure if many people do it, but that would definitely clean up what we do.

PR #159, opened 2019-01-18
add support for the java.time.Duration type

massimosiani:
This adds the converters for the java.time.Duration for the Mongo driver.


DanielFran:
Thans @massimosiani

PR #166, opened 2019-01-31
Upgrade jaxb-impl version to 2.3.2

DanielFran:


DanielFran:
@jdubois added it in commit https://github.com/jhipster/jhipster/commit/42815832b63d2d1446bed96c878c7a785bbaa3e7 when migrating from JDK8 to JDK11 probably because he had error ClassNotFoundException missing jaxb

PR #171, opened 2019-02-07
add PageUtil interface with utility method

Gigaxel:
Issue: jhipster/generator-jhipster#9209

add PageUtil interface with method for creating a page from a list

resolve jhipster/generator-jhipster#9209


DanielFran:
Thanks @Gigaxel

PR #182, opened 2019-02-13
Add copy method and constructor

Blackdread:
See issue [9084](https://github.com/jhipster/generator-jhipster/issues/9084)


gzsombor:
Nice, I like the idea!

PR #196, opened 2019-02-21
Removed HTTP2 useUndertowUserCipherSuitesOrder property

DanielFran:
See https://github.com/jhipster/generator-jhipster/pull/9300

cc @PierreBesson


jhipster-framework/src/main/java/io/github/jhipster/security/ssl/UndertowSSLConfiguration.java
```diff
@@ -37,7 +36,7 @@
  * This can only be activated with HTTPS and a cipher suite defined by the user (server.ssl.ciphers).
  * <p>
  * Please note that when using JHipster, you can use the `server.ssl.ciphers` property that is commented out
```
PierreBesson: remove the "is commented out" part as it is not commented out anymore.

DanielFran: Done

PR #202, opened 2019-03-06
Use UriComponentsBuilder instead of HttpServletRequest in generatePag…

cbornet:
…inationHttpHeaders

It can then be used in resources like this
```java
    @GetMapping("/users")
    public ResponseEntity<List<UserDTO>> getAllUsers(@RequestParam MultiValueMap<String, String> queryParams, UriComponentsBuilder uriBuilder, Pageable pageable) {
        final Page<UserDTO> page = userService.getAllManagedUsers(pageable);
        HttpHeaders headers = PaginationUtil.generatePaginationHttpHeaders(uriBuilder.queryParams(queryParams), page);
        return new ResponseEntity<>(page.getContent(), headers, HttpStatus.OK);
    }
```
This code works for both MVC and Webflux.

### DIFF
--- a/jhipster-framework/src/test/java/io/github/jhipster/web/util/PaginationUtilUnitTest.java
+++ b/jhipster-framework/src/test/java/io/github/jhipster/web/util/PaginationUtilUnitTest.java
@@ -24,12 +24,13 @@ import org.springframework.data.domain.Page;
 import org.springframework.data.domain.PageImpl;
 import org.springframework.data.domain.PageRequest;
 import org.springframework.http.HttpHeaders;
-import org.springframework.mock.web.MockHttpServletRequest;
+import org.springframework.web.util.UriComponentsBuilder;
 
 import java.util.ArrayList;
 import java.util.List;
 
-import static org.junit.Assert.*;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
 
 /**
  * Tests based on parsing algorithm in app/components/util/pagination-util.service.js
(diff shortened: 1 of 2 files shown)


jhipster-framework/src/test/java/io/github/jhipster/web/util/PaginationUtilUnitTest.java
```diff
@@ -40,171 +41,142 @@
...
+        String expectedData = "</api/_search/example?query=Test%3E%3Btest&page=0&size=0>; rel=\"last\"," +
+            "</api/_search/example?query=Test%3E%3Btest&page=0&size=0>; rel=\"first\"";
         assertEquals(expectedData, headerData);
         List<String> xTotalCountHeaders = headers.get("X-Total-Count");
-        assertTrue(xTotalCountHeaders.size() == 1);
-        assertTrue(Long.valueOf(xTotalCountHeaders.get(0)).equals(0L));
-    }
-
-    @Test
-    public void sortAndCriteriaTest() {
```
cbornet: I didn't see the added value of this test so I removed it.

cbornet:
See https://github.com/jhipster/generator-jhipster/pull/9379

jhipster-framework/src/test/java/io/github/jhipster/web/util/PaginationUtilUnitTest.java
jdubois: this is to test the bug we had previously, you were losing the sort and criteria when navigating to another page

cbornet: Yes but the fix uses a generic way to add the query params. So it's already tested in the previous tests

pvliss:
@jdubois + @DanielFran Is it ok with you guys if I merge this? LGTM and is required to get the CI back to green status. Current approach with `HttpServletRequest` is not working with webflux anyways. Once this is merged we can relaunch the CI build at jhipster/generator-jhipster#9379 to see if everything is working properly again. WDYT?

cbornet:
Yes, this should fix the CI so let's merge.
@jdubois can you push the SNAPSHOT ?

PR #230, opened 2019-04-03
Fix info contributor conditional property

vishal423:
Fix conditional property that allows you to enable/disable this info contributor via spring boot configurations


PierreBesson:
Thanks for the fix ! There might have been some confusion between @ConditionalOnProperty and @ ConditionalOnEnabledInfoContributor.

PR #252, opened 2019-04-25
Review Sonar support

DanielFran:


DanielFran:
@pascalgrimaud @jdubois Should we add this project also in https://sonarcloud.io/organizations/jhipster/projects ?

pascalgrimaud:
It would be nice, but I'm afraid about the result :fearful:

DanielFran:
And https://sonarcloud.io/dashboard?id=io.github.jhipster.sample%3Ajhipster-sample-application should be moved also to the organization?
And what about adding all samples?

pascalgrimaud:
About moving to organization: yeah, something I had in mind since months
About all samples: we need to think about that, as other samples can changed regularly, by adding new entities in CI, changing options, etc. But sure, it will help to detect bad code

PR #299, opened 2019-06-19
Fixing generator-jhipster#9836 

gzsombor:
'greaterOrEqualThan' should be 'greaterThanOrEqual'


gzsombor:
The build failed in the test phase, as it is an incompatible change ...

gzsombor:
Following @Blackdread suggestions, I've re-added the original, badly worded setters as deprecated code - but not the getters. This way, if someone upgrades their app, they will get Java compilation errors if they used the old getters, but the javascript client code could still work as is - we can remove the deprecated code in the next major release.

jdubois:
OK, thanks a lot @gzsombor !

PR #324, opened 2019-07-08
Redis cache support

Shaolans:
This PR add the maven dependency for redis cache and update the JHipsterProperties to support Redis.
Related to https://github.com/jhipster/generator-jhipster/pull/10057
https://github.com/jhipster/generator-jhipster/issues/9280

### DIFF
--- a/jhipster-dependencies/pom.xml
+++ b/jhipster-dependencies/pom.xml
@@ -81,6 +81,7 @@
         <testcontainers.version>1.11.3</testcontainers.version>
         <xmemcached.version>2.4.6</xmemcached.version>
         <xmemcached-provider.version>4.1.3</xmemcached-provider.version>
+        <redisson.version>3.11.1</redisson.version>
     </properties>
 
     <dependencyManagement>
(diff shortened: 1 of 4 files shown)


jhipster-dependencies/pom.xml
```diff
@@ -81,6 +81,7 @@
         <testcontainers.version>1.11.3</testcontainers.version>
         <xmemcached.version>2.4.6</xmemcached.version>
         <xmemcached-provider.version>4.1.3</xmemcached-provider.version>
+        <redisson.version>3.11.1</redisson.version>
```
DanielFran: @Shaolans Please move this property after <prometheus-simpleclient.version>, properties should be by default ordered by name

Shaolans: Done thanks !

DanielFran:
@Shaolans you can merge if you think it is ready at least in this project.

Shaolans:
@DanielFran I do not have the right for merging, but I think the PR on the generator should be approved first to ensure everything is okay :+1:

PR #371, opened 2019-08-16
add spring boot configuration processor

PierreBesson:
Thi enable completion of jhipster properties in IDE

Fix jhipster/generator-jhipster/issues/10195

In future version of the lib, you will have this.
Screenshot from 2019-08-16 10-34-47

Screenshot from 2019-08-16 10-36-34


pascalgrimaud:
The current build in master is already broken for couchbase, so it's not related to this PR. That's why I'm merging this

PR #379, opened 2019-08-23
Creating application properties within the jhipster framework

SudharakaP:
Related to https://github.com/jhipster/generator-jhipster/pull/10262#pullrequestreview-278878151


SudharakaP:
The three tests failing here seems not related to the issue; could someone please restart them in order to verify :cry:

PR #383, opened 2019-08-29
Add Support for Caffeine

SudharakaP:
This replaces Ehcache with Caffeine; related to issue; https://github.com/jhipster/generator-jhipster/issues/9351.

### DIFF
--- a/jhipster-framework/src/main/java/io/github/jhipster/config/JHipsterDefaults.java
+++ b/jhipster-framework/src/main/java/io/github/jhipster/config/JHipsterDefaults.java
@@ -64,6 +64,12 @@ public interface JHipsterDefaults {
             long maxEntries = 100;
         }
 
+        interface Ehcache {
+
+            int timeToLiveSeconds = 3600; // 1 hour
+            long maxEntries = 100;
+        }
+
         interface Infinispan {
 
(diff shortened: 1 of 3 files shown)


pmverma:
The title feels like we are going to drop Ehcache support in favour of Caffeine. Can you revise it?

SudharakaP:
@pmverma : Actually this is exactly what I did; in the issue; https://github.com/jhipster/generator-jhipster/issues/9351 it is mentioned to make Caffeine the "default" cache and so I thought that Caffeine should replace Ehcache as its more performant in every sense. 

But you are right maybe the point is to add Caffeine as another option @jdubois ? :fearful:

pmverma:
Usually this kind of major modifications goes through Google group mail discussion.

You can also see the statistics for last year at https://start.jhipster.tech/#/statistics,
Ehcache is at first place.
Moreover, the ehcache people have contributed as well to have a better integration with JHipster.

DanielFran:
@SudharakaP From what I understaood, it was only to add new cache option implementation and maybe define by default caffeine.
I do not believe that it was do delete ehcache implementation!

jdubois:
Indeed, I would like to keep ehcache by default (great contribution by the ehcache team, and also it's widely used and known), and have Caffeine as an option

PR #428, opened 2019-10-08
Fix builds, using JDK11

pascalgrimaud:


DanielFran:
@pascalgrimaud Thanks for the changes.
Should we consider to switch from Travis to Github actions like in the generator project?

pascalgrimaud:
Yes, it's the plan. But not enough time these last weeks :(

pascalgrimaud:
@DanielFran : I'm merging this, so I can start the release for the jhipster lib

pascalgrimaud:
I can't do a release right now, as https://oss.sonatype.org is down

PR #506, opened 2019-12-17
Add doc how to install SNAPSHOT version

kaidohallik:


jdubois:
We should also publish one snapshot automatically, so this issue would be solved
cc @pascalgrimaud we need to work on this, I got several ideas

pascalgrimaud:
yes, first let's migrate to GitHub Actions, then we can think about releasing snapshot :)

jdubois:
@pascalgrimaud look at the dev mailing list, I just did it 😀

PR #507, opened 2019-12-17
Allow any characters in notification parameter

kaidohallik:
This is the first part for the jhipster/generator-jhipster#10877

This alert parameter is used:
* in Angular user management and entities
* in React user management and entities
* in Vue.js user management only (in entities instead of using HTTP header returned from JHipster server side library, this parameter is used from data already available in client application)

Currently in symbols used by entity `id` or user `login`, symbol @ causes incompatibility between old and new version, because `@` is encoded to `%40`.

So to avoid showing `@` as `%40` in default generated projects, blueprints should adapt this change.

I'll do PR for Angular and React in the `generator-jhipster` and in `vue.js` blueprint to adapt this change there.


kaidohallik:
Client side PR-s:
* Angular and React: jhipster/generator-jhipster#10972
* Vue.js: jhipster/jhipster-vuejs#514

PR #509, opened 2019-12-18
Build instructions for Windows to skip signing artifacts

gmarziou:
In Windows Powershell, running `.\mvnw.cmd clean install -Dgpg.skip=true`  fails as:

> [ERROR] Unknown lifecycle phase ".skip=true". You must specify a valid lifecycle phase or a goal in the format

The correct way is to use quotes  `.\mvnw.cmd clean install -D"gpg.skip=true"`


gmarziou:
I'll do it also.

kaidohallik:
Thanks @gmarziou for fixing, thanks @DanielFran for merging!

PR #525, opened 2020-01-02
Sonar

DanielFran:


pascalgrimaud:
@DanielFran : just for my information, what are you trying to fix plz ? is there any issue with current Sonar Analysis ?

DanielFran:
@pascalgrimaud The current approach works in fact, but since this is a multi-module project I was trying to implement it correctly.
Because I do not have access to the sonar token, I can just see the result comitting here...

SudharakaP:
@DanielFran : Just throwing an idea. 😄 When I was making some changes related to this Sonar analysis recently, I noticed that JaCoCo has a goal for multi module projects which we might want to use; 

References: 

https://github.com/jacoco/jacoco/wiki/MavenMultiModule

https://www.eclemma.org/jacoco/trunk/doc/report-aggregate-mojo.html

DanielFran:
@SudharakaP That was the one I was trying to use... but there is 3 different options, now I was planning to use the last one, create a specific module for reporting that seems the best option.

PR #539, opened 2020-01-13
Use the JHipster BOM to manage the Azure Spring Cloud client version

jdubois:


jdubois:
Thanks! It's not used yet (the version is fixed at the moment), so let's merge as there is no risk

PR #561, opened 2020-02-02
add neo4j dependency versions

atomfrede:
In preparation for the upcoming neo4j support these two deps are needed and it would be nice to have these versions managed.


DanielFran:
@atomfrede Looking at https://github.com/jhipster/generator-jhipster/pull/11226, this seems not needed any other additional dependency.
This PR is ready to be merged?

atomfrede:
Yes we can merge this. just adding versions and dependencies here.

PR #608, opened 2020-03-02
Reactive version of ResponseUtil

murdos:
This is needed for https://github.com/jhipster/generator-jhipster/pull/11416 and https://github.com/jhipster/generator-jhipster/issues/11404

### DIFF
--- a/jhipster-framework/pom.xml
+++ b/jhipster-framework/pom.xml
@@ -111,6 +111,11 @@
         </dependency>
         <dependency>
             <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-webflux</artifactId>
+            <optional>true</optional>
+        </dependency>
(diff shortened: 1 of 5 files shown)


DanielFran:
@murdos Why create a specific ResponseUtil reactive classe and not reuse the existing?

murdos:
@DanielFran : a non reactive JHipster project can't use an utility class that depends on Reactor, because it doesn't have the dependency. Thus it was necessary to use a distinct class.

DanielFran:
Yes i saw it 😅😂🥱😴

pascalgrimaud:
I plan to do a new release of jhipster dependencies (3.6.0)
Do you want to merge this @murdos ?

murdos:
Yes, that would be nice, if there's no objection!

PR #628, opened 2020-03-17
Offload blocking CSRF token generation to boundedElastic scheduler

cbornet:
See https://github.com/jhipster/generator-jhipster/pull/11009

### DIFF
--- a/jhipster-framework/src/main/java/io/github/jhipster/web/filter/reactive/CookieCsrfFilter.java
+++ b/jhipster-framework/src/main/java/io/github/jhipster/web/filter/reactive/CookieCsrfFilter.java
@@ -1,6 +1,5 @@
 package io.github.jhipster.web.filter.reactive;
 
-
 import org.springframework.http.ResponseCookie;
 import org.springframework.http.server.reactive.ServerHttpRequest;
 import org.springframework.security.web.server.csrf.CsrfToken;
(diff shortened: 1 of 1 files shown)


jhipster-framework/src/main/java/io/github/jhipster/web/filter/reactive/CookieCsrfFilter.java
```diff
@@ -30,16 +30,18 @@
...
-                .httpOnly(false)
-                .path(getRequestContext(exchange.getRequest()))
-                .secure(Optional.ofNullable(exchange.getRequest().getSslInfo()).isPresent())
-                .build();
-            exchange.getResponse().getCookies().add(CSRF_COOKIE_NAME, cookie);
-
-        }).then(Mono.defer(() -> chain.filter(exchange)));
+        return csrfToken
+            .publishOn(Schedulers.boundedElastic())
+            .doOnSuccess(token -> {
```
vishal423: As per my understanding, we only need to subscribe to get the CSRF cookie on the client side something like [this](https://github.com/vishal423/ngx-hipster/blob/fdc20e7c005348d314e580f550107bf21d662d76/demo/ngx-hipster-sample-api/src/main/java/io/github/vishal423/ngxhipster/config/SecurityConfig.java#L42)

cbornet: @vishal423 could you try to improve it in a PR ? The problem of the token generation being blocking will remain so you need to use an appropriate scheduler

PR #657, opened 2020-04-07
Add a specific profile for Azure

jdubois:
Like we already have for Heroku or AWS


jdubois:
Is the CI falky? I don't think this change could break some of those builds (or it would have broken ALL builds)

DanielFran:
Yes we have some issue with webflux

PR #672, opened 2020-04-21
Publish artifacts to GitHub Packages

murdos:
New workflow that publishes all push to master to Github Packages.

I had to disable the generation and deployment of javadoc and sources since it's currently not supported by GitHub Packages (it generates an http 400 error when you then try retrieve metadata).

### DIFF
--- a/.github/workflows/gh-packages.yml
+++ b/.github/workflows/gh-packages.yml
@@ -1,0 +1,36 @@
+#
+# Copyright 2016-2020 the original author or authors from the JHipster project.
+#
+# This file is part of the JHipster project, see https://www.jhipster.tech/
+# for more information.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


murdos:
@jdubois : I've conditionally included https://oss.sonatype.org/content/repositories/snapshots in the generated build config (Maven and Gradle): https://github.com/jhipster/generator-jhipster/pull/11598 
This will be great for people using master of generator-jhipster.

The second reason I wanted to publish artifacts after each merge to master was to skip the building of jhipster lib during generator-jhipster CI tests. So I guess it won't be possible with only nightly publications.

jdubois:
Indeed you will have to wait for the next day... The best solution would be to have some kind of "jhipster" credentials (=not an individual's credentials, but an org's credential).
Maven Central is really too limited for this, a better option would be to switch to JFrog's artifactory. I've never had the time to learn it well enough, but I believe it's the best solution.

PR #678, opened 2020-04-25
Upgrade Couchmove to 2.0

tchlyah:
Add Full Text Search Option (see https://github.com/jhipster/generator-jhipster/pull/11421)

### DIFF
--- a/jhipster-dependencies/pom.xml
+++ b/jhipster-dependencies/pom.xml
@@ -45,7 +45,7 @@
         <bucket4j.version>4.10.0</bucket4j.version>
         <!-- see https://github.com/jhipster/generator-jhipster/issues/11169 -->
         <cassandra-unit-spring.version>2.2.2.1</cassandra-unit-spring.version>
-        <couchmove.version>1.1</couchmove.version>
+        <couchmove.version>2.0</couchmove.version>
         <!-- waiting for https://github.com/jhipster/generator-jhipster/issues/11491 -->
         <couchbase-client.version>2.7.13</couchbase-client.version>
         <couchbase-encryption.version>2.0.1</couchbase-encryption.version>


DanielFran:
@tchlyah Maybe you can review couchbase-client version?

mraible:
This seems to cause a compilation failure in couchbase + elasticsearch + oauth: https://github.com/jhipster/generator-jhipster/pull/11623/checks?check_run_id=618427010#step:14:53

mraible:
Might be related: `org.apache.commons.codec.binary.Base64` is no longer available. https://github.com/jhipster/generator-jhipster/pull/11623/checks?check_run_id=618427114#step:14:185

mraible:
I've confirmed that changing the `webflux-couchbase` project to use v1.1 fixes compilation:

```xml
<dependency>
    <groupId>com.github.differentway</groupId>
    <artifactId>couchmove</artifactId>
    <version>1.1</version>
</dependency>
```

Therefore, I'm reverting this change.

tchlyah:
sorry @mraible I didn't saw your message soon enough. This update is meant for the Couchbase Full Text Search option, and https://github.com/jhipster/generator-jhipster/pull/11421 fixes the issue.

I don't think that reverting this is a good solution since we are using here a transitive dependency and we shouldn't rely on it, it's better to add directly the dependency as we do for cassandra. This what I did here https://github.com/jhipster/generator-jhipster/pull/11421/commits/a730bd20dbe58f3f2367612309faf66fa9d4c929.

I can separate the fix from the FTS PR if you don't want to wait for its review.

murdos:
> I can separate the fix from the FTS PR if you don't want to wait for its review.

Yes, please do so, because currently `master` is broken.

tchlyah:
Done in https://github.com/jhipster/generator-jhipster/pull/11640, it should be ok

tchlyah:
> @tchlyah Maybe you can review couchbase-client version?

Done in https://github.com/jhipster/jhipster/pull/680

PR #751, opened 2020-07-14
Added vue workflow

DanielFran:


pascalgrimaud:
You're faster than me :-)
I plan to work on all builds, to reduce the number in main generator-jhipster and align it there.
But before doing it, we need to discuss about options to keep and options to remove.

But it's OK to merge it now.

PR #752, opened 2020-07-18
Delete deprecated methods

DanielFran:


murdos:
Since with since changes master includes breaking changes, a maintenance branch for 6.x should probably be created if we need to do new 6.x releases.

pascalgrimaud:
yes you're right @murdos : if needed, I'll create a separate branch from the last tag

PR #764, opened 2020-07-21
Upgrade Sprinfox to v3.0.0

cbornet:
See https://github.com/jhipster/generator-jhipster/issues/12125

* Generates OAS  on /v3/api-docs
* Generates Swagger2 on /v2/api-docs for backward compatibility
* Some breaking changes:
  * `swagger` maven profile becomes `api-docs`
  * `jhipster.swagger` property becomes `jhipster.api-docs`
  * `SwaggerCustomizer` becomes `SpringfoxCustomizer`
  * `swaggerSpringfoxApiDocket` becomes `openapiSpringfoxApiDocket`
  * `swaggerSpringfoxManagementDocket`becomes `openAPISpringfoxManagementDocket`
* Unit tests on the autoconfiguration are replaced by more relevant integration tests
* A `jhipster.servers` is introduced (replaces host+protocols in OpenAPI v3) but there seems to be a bug in Springfox


jhipster-framework/src/main/java/io/github/jhipster/config/apidoc/SpringfoxAutoConfiguration.java
```diff
@@ -121,28 +122,28 @@ public Docket swaggerSpringfoxApiDocket(List<SwaggerCustomizer> swaggerCustomize
...
-     * JHipster Swagger Customizer
+     * JHipster Springfox Customizer
      *
-     * @return the Swagger Customizer of JHipster
+     * @return the Sringfox Customizer of JHipster
      */
     @Bean
-    public JHipsterSwaggerCustomizer jHipsterSwaggerCustomizer() {
-        return new JHipsterSwaggerCustomizer(properties);
+    public JHipsteSpringfoxCustomizer jHipsterSpringfoxCustomizer() {
```
pascalgrimaud, line 130: Typo: JHipsteSpring -> JHipsterSpring

pascalgrimaud:
@cbornet : what do you think about changing profile `swagger` to `openapi` (instead of `api-docs`), as it would be simpler to remember

cbornet:
I had done that at first. But then I thought it could be confusing with the API-first option and the API client generation which both also use OpenAPI. The advantage of api-docs is that the target is clear. And we also use apidoc in the config package name.

DanielFran:
@cbornet Just waiting that you fix the typo....

cbornet:
Typo fixed.
Is everyone OK with `api-docs` (see my comment) or do I rename to `openapi` ? @jhipster/developers

pascalgrimaud:
I'm ok with both :) choose the one you prefer

DanielFran:
@cbornet Feel free to merge it.

cbornet:
Once it's merged, #12133 needs to be merged quickly as there are some breaking changes. @DanielFran could you review it also ?

PR #785, opened 2020-08-16
Deploy snapshots to GitHub packages.

mshima:
Related to https://github.com/jhipster/generator-jhipster/issues/12251

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,7 +17,7 @@
 # limitations under the License.
 #
 
-name: Sonar analysis
+name: Test, Sonar and deploy.
 
 on:
   push:
@@ -26,6 +26,7 @@ on:
 
 jobs:
     build-and-analyse:
+      if: github.repository == 'jhipster/jhipster'
       runs-on: ubuntu-latest
       steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


DanielFran:
@jdubois We already discuss to deploy snapshots. Now that we can on GitHub it is the opportunity to do it...

pascalgrimaud:
As mentioned in PR and at https://github.com/jhipster/generator-jhipster/issues/12251#issuecomment-674552308

I'm not totally convinced about this approach. But let's discuss

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,47 @@
...
+        with:
+          java-version: 11
+      - name: Cache Maven archetypes
+        uses: actions/cache@v2
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-maven-
+      - name: Deploy to github packages
```
pascalgrimaud, line 44: I propose some change:
- rename `sonar.yml` to something else, like `build.yml`
- add this new part in `build.yml` so we'll deploy only after mvn test and mvn sonar

pascalgrimaud:
After thinking about it, we can give a try and deploy to GitHub package.
It won't have any impact directly, excepting if we decide to use `release` value + snapshot version.

pascalgrimaud:
OK then, so let's give a try !
I'm merging this

PR #827, opened 2020-09-10
Update to Cucumber 6

murdos:
See https://github.com/jhipster/generator-jhipster/issues/11136


murdos:
@DanielFran : this will fail again, because I've not switched to generator branch here, and I don't think it's necessary. I've done it on generator side, and tests are passing.

DanielFran:
@murdos Was just checking if errors were due to an old run... and since it was green in generator was expecting to validate it

DanielFran:
So this seems ready. Do you want I switch on both project to master branch and merge it?

murdos:
> 
> 
> So this seems ready. Do you want I switch on both project to master branch and merge it?

Thanks, but I'll do it right now ;)

